PR #41, opened 2014-12-03
AWS Provisioner does not work correctly with Machine_batch

lynchc:
If you create a series of machines surrounded with a machine_batch, they will not be created in parallel. The allocate_machine method will be executed from multiple threads and will fail with Argument errors on actual_instance.exists?

A quick solution was to add in a mutex to the method. It would still not be ran in parallel but would create safety to avoid failures.

After discussion with @johnewart, he decided that the best solution was to have the aws driver implement it's own version of allocate_machines that accepts and uses the cheffs parallelizer correctly

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -228,6 +228,15 @@ module AWSDriver
       end
     end
 
+    def allocate_machines(action_handler, specs_and_options, parallelizer)
+      #Chef::Log.warn("#{specs_and_options}")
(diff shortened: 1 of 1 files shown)


lynchc:
This is the link to the aws discussion explaining how the ruby sdk isn't thread safe and how to circumvent:
https://forums.aws.amazon.com/thread.jspa?threadID=100212

jkeiser:
:+1: looks just like the fog one.

jkeiser:
Checked in as a4d6f75

lynchc:
The best ideas are the ones you steal from people who are smarter than you

PR #47, opened 2014-12-16
Set image_id as a bootstrap_option.

### LINKED ISSUE #46: image_id should be a bootstrap_options instead of machine_option?
As far as I can see, the image_id should be a bootstrap_option instead of a machine_option as described [here](https://github.com/opscode/chef-provisioning/blob/master/docs%2Fproviders%2Faws.md).

This is also to be coherent with the fog:AWS driver.

miguelcnf:
This is to be coherent with the docs and the fog:AWS driver.

Should fix #46.

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -212,16 +212,15 @@ module AWSDriver
     def allocate_machine(action_handler, machine_spec, machine_options)
       actual_instance = instance_for(machine_spec)
       if actual_instance == nil || !actual_instance.exists? || actual_instance.status == :terminated
-        image_id = machine_options[:image_id] || default_ami_for_region(@region)
         bootstrap_options = machine_options[:bootstrap_options] || {}
-        bootstrap_options[:image_id] = image_id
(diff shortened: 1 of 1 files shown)


jkeiser:
I like this :)  I've been wondering why we do this the way we do.

This may be a breaking change for people, but I think putting :image_id in the machine_options is the wrong thing, and your patch is right.

We need to update the README before checking this in, and note the breaking change as well (since `:image_id` in `machine_options` won't be supported anymore if I read it right).

miguelcnf:
Yea and as I mentioned that would come aligned with the fog:AWS driver, otherwise it might be confusing to people, I know it was for me! :)

@jkeiser not sure which README you're referring to, is it the chef-provisioning one? Happy to update the docs if you can point me to it.

christinedraper:
BTW, this is more than being consistent. Without it, the following doesnt work:

machine_image "appserver_image"
end

machine "appserver"
from_image "appserver_image"
end

jkeiser:
We merged #89 due to less breaking changes, but a breaking change with a deprecation error and a README fix are both welcome :)

Issue #42, opened 2014-12-10
driver image methods are empty


miguelcnf:
Just leaving this here for tracking purposes. cc/ @jkeiser

The image methods used by the machine_image resources are implemented but empty:

```ruby
    # Image methods
    def allocate_image(action_handler, image_spec, image_options, machine_spec)
    end

    def ready_image(action_handler, image_spec, image_options)
    end

    def destroy_image(action_handler, image_spec, image_options)
    end
```

This causes that a chef-client run using the machine_image resource with the aws driver not to produce the desired results but also not failing.

miguelcnf:
This was closed by merging #54.

PR #54, opened 2014-12-18
Initial commit of missing image methods.

miguelcnf:
This tries to follow the same implementation logic of the machine methods and backports
the actions from the fog:AWS driver.

This should be enough to solve #42.


jkeiser:
:+1:.  I'm sad we have to duplicate all this progress logic.  Will have to get better :)

jkeiser:
Committed as 8cb1640

miguelcnf:
Thanks - this was really a blocker for me to move on from fog:AWS to native AWS! :)

Issue #51, opened 2014-12-17
Ugly output message during machine resource convergence


tpetchel:
- It appears that this message is missing a few words (between 'username' and 'and' in "with SSH username  and machine_options" and also "Using .")
- I'm not sure what the actual problem is that this is describing. 
- It would be difficult for most users (IMO) to know what it means to change `chef_provisioning.location.ssh_username`.

```
* machine[webserver-1] action converge[2014-12-16T15:57:53-08:00] WARN: Server webserver-1 was 
created with SSH username  and machine_options specifies username root.  Using .  Please edit the 
node and change the chef_provisioning.location.ssh_username attribute if you want to change it.
```

jkeiser:
This is more about the error message, honestly.  Probably still open.

randomcamel:
The fix for this should probably be ported to `chef-provisioning-fog`, which has the same error message [here](https://github.com/chef/chef-provisioning-fog/blob/master/lib%2Fchef%2Fprovisioning%2Ffog_driver%2Fdriver.rb#L677).

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

Issue #44, opened 2014-12-16
Can't specify ssh_username via machine_options


miguelcnf:
Because of:

```ruby
          machine_spec.location = {
              'driver_url' => driver_url,
              'driver_version' => Chef::Provisioning::AWSDriver::VERSION,
              'allocated_at' => Time.now.utc.to_s,
              'host_node' => action_handler.host_node,
              'image_id' => machine_options[:image_id],
              'instance_id' => instance.id
          }
```

And:

```ruby
          username = machine_spec.location['ssh_username'] || default_ssh_username
```

We can't set ssh_username as a machine_option.

jkeiser:
This should be fixed with the merge of #65, would you mind retesting when you get a chance?

jkeiser:
I'm going to close in the meantime, but @miguelcnf please do reopen if #65 didn't fix it--I know you had a different fix in the offing.

PR #55, opened 2014-12-19
Use machine_options[:ssh_username] for transport.

miguelcnf:
Give precedence to machine_options instead of machine_spec when itcomes
choosing the ssh username.

Also update the machine_spec with the machine_options value if transport
is successful.

This replaces #45 with a clean commit history.

Should solve #44 and #51.

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -253,9 +253,9 @@ module AWSDriver
           end
         end
         wait_until_ready(action_handler, machine_spec, instance)
-        wait_for_transport(action_handler, machine_spec, machine_options)
       end
 
+      wait_for_transport(action_handler, machine_spec, machine_options)
       machine_for(machine_spec, machine_options, instance)
 
     end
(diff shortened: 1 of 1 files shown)


jkeiser:
In general, we haven't been giving precedence to machine_options because keys can drift over time; but I'm leaving this open even though #65 was committed, because there is a real discussion to have on this point.  I'm not happy at all with our credentials story, yet.

jkeiser:
@miguelcnf for the moment I'm moving this one to Later since the primary issue (`ssh_username` is not honored) is fixed; but I still want to have this discussion.  It seems to me that there are real tradeoffs either way--if you prefer machine_options, it becomes changeable (which is super good), and if you prefer machine_spec, then recipe drift won't break your ability to reach older machines (which is also good).

jkeiser:
Closing this for now; we're still aware of it but not planning to merge this at the moment.

poliva83:
@jkeiser Can we reopen this discussion? 

`chef-provisioning-aws` today has implemented ssh transport (https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/driver.rb#L948) in a way you can't update `ssh_username` on second convergence of machine node unless you actually delete the machine first. This is because machine_spec is higher precedence then what you set in machine_options. There seems to be no workaround is if you desire to update ssh_username (and force 2nd chef-client run) without deleting your machine.

I'd like to have this converstation because we followed this same implementation for our [chef-provisioning-opennebula](https://github.com/blackberry/chef-provisioning-opennebula) driver. I don't really want to our driver to have different logic then `chef-provisioning-aws` since its the most popular driver.

miguelcnf:
When I first wrote this I found it annoying but not a show stopper, I still look at it the same but in my head it makes total sense that machine_options have priority over machine_spec exactly for what @poliva83 described.

@tyler-ball let me know if you want me to look at this branch again and resolve conflicts etc.

bbuczynski:
a use case to have machine_spec with higher precedence than machine_options would be if there is a single provisioning user that can deploy VMs (for example to prod) and nobody else should have access to those machines ... how about making it configurable? this is how I implemented unique VM names, fqdn enforcement, and stop method override in opennebula driver

tyler-ball:
I don't think we should make the machine_options <=> machine_spec precedence configurable.  It seems like adding a knob where we don't need one.

I think that `machine_options` should override `machine_spec`.  It would be a breaking change so we would need a major version bump, but after documenting it I think it would make more sense to users.  What you specify in your recipe should always take precedence over what has been stored in the background.

cgunasree08:

PR #56, opened 2014-12-19
Provide methods to create and delete VPC Subnets

meekmichael:
A provider for provisioning subnets within an AWS VPC Environment.

```
  subnet "provisioning-vpc-subnet-a" do
    cidr_block "10.0.1.0/26"
    vpc "provisioning-vpc"
    availability_zone "eu-west-1a"
  end
```

Omitting `availablity zone` would have Amazon pick the AZ for you.

### DIFF
--- a/lib/chef/provider/aws_subnet.rb
+++ b/lib/chef/provider/aws_subnet.rb
@@ -1,0 +1,52 @@
+require 'chef/provider/aws_provider'
+require 'date'
+
+class Chef::Provider::AwsSubnet < Chef::Provider::AwsProvider
+
+  action :create do
+    fail "Can't create a Subnet without a CIDR block" if new_resource.cidr_block.nil?
+
+    if existing_subnet == nil
+      converge_by "Creating new Subnet #{fqn} in VPC #{new_resource.vpc} in #{new_resource.region_name}" do
+        opts = { :vpc => vpc_id }
+	opts[:availability_zone] = new_resource.availability_zone if new_resource.availability_zone
+        subnet = ec2.subnets.create(new_resource.cidr_block, opts)
+        subnet.tags['Name'] = new_resource.name
+        subnet.tags['VPC'] = new_resource.vpc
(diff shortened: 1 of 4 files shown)


erikvanbrakel:
I'm using this now, seems to work fine for the basic creation stuff. Any timelines on this? Or is there anything I can contribute to speed this up?

jkeiser:
I love that this is using tags for idempotence!  @jdmundrawala would you mind taking a quick look at this one as well, for sanity on the definition?

jkeiser:
Given what I see here, this looks spot on: http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/EC2/SubnetCollection.html#create-instance_method

jkeiser:
Merged as 844b25c (though @jdmundrawala's sanity check would still be super useful :)

lib/chef/provider/aws_subnet.rb
```diff
@@ -0,0 +1,52 @@
...
+    @vpc_id ||= begin
+      ec2.vpcs.with_tag('Name', new_resource.vpc).first
+    rescue
+      nil
+    end
+  end
+
+  def existing_subnet
+      @subnet_id ||= begin
+      ec2.subnets.with_tag('Name', new_resource.name).first
```
jdmundrawala: If there were multiple of these, would a warning be in order?

lib/chef/provider/aws_subnet.rb
```diff
@@ -0,0 +1,52 @@
...
+        existing_subnet.delete
+      end
+    end
+
+    new_resource.delete
+  end
+
+  def vpc_id
+    @vpc_id ||= begin
+      ec2.vpcs.with_tag('Name', new_resource.vpc).first
```
jdmundrawala: Same comment as from existing_subnet

jdmundrawala:
:+1: on the use of tags. We might want to consider namespacing them. Also, we should have a task to figure out where else it is appropriate to do this.

PR #69, opened 2015-01-27
fix for chef/chef-provisioning/aws#68

lynchc:


lynchc:
whoops. I'll fix after lunch

Issue #68, opened 2015-01-26
Error when converging cookbooks depending on 'aws' cookbook


esciara:
Whenever a cookbook using the chef-provisioning-aws driver depends on the ['aws' cookbook](https://github.com/opscode-cookbooks/aws), converging fails at cookbooks compilation time with the following error:
```
[..]
Compiling Cookbooks...

================================================================================
Recipe Compile Error in /path/to/your/cookbook/development/directory/.chef/local-mode-cache/cache/cookbooks/aws-bug-cookbook/recipes/default.rb
================================================================================

TypeError
---------
superclass mismatch for class AwsEbsVolume

Cookbook Trace:
---------------
  /path/to/your/cookbook/development/directory/.chef/local-mode-cache/cache/cookbooks/aws-bug-cookbook/recipes/default.rb:15:in `require'
  /path/to/your/cookbook/development/directory/.chef/local-mode-cache/cache/cookbooks/aws-bug-cookbook/recipes/default.rb:15:in `from_file'

Relevant File Content:
----------------------
/path/to/your/cookbook/development/directory/.chef/local-mode-cache/cache/cookbooks/aws-bug-cookbook/recipes/default.rb:

  8:  #
  9:  # Licensed under the Apache License, Version 2.0 (the "License");
 10:  # you may not use this file except in compliance with the License.
 11:  # You may obtain a copy of the License at
 12:  #
 13:  #     http://www.apache.org/licenses/LICENSE-2.0
 14:  
 15>> require 'chef/provisioning/aws_driver'
 16:  with_driver 'aws'
 17:  
[..]
```
Full instructions on https://gist.github.com/esciara/d0921bded64568dc6424/edit

This stops me completely from deploying cookbooks such as https://github.com/bflad/chef-confluence .

Jimflip:
That gist is missing

esciara:
@Jimflip sorry... had left a '/edit' in the link. Corrected now, you should be able to access it

pburkholder:
The `aws` LWRP is going to create these resources:

    Chef::Resource::AwsEbsRaid             
    Chef::Resource::AwsEbsVolume           
    Chef::Resource::AwsElasticIp           
    Chef::Resource::AwsElasticLb           
    Chef::Resource::AwsInstanceMonitoring  
    Chef::Resource::AwsResourceTag

of which AwsEbsVolume is going to have the superclass conflict. Not sure how to work around that.

esciara:
[...]

PR #70, opened 2015-01-27
changed class ebs volume class name: AwsEbsVol to avoid naming conflict

lynchc:


esciara:
This would correct problem described in #68

tyler-ball:
I'm not a big fan of this solution because it doesn't mitigate any future name collisions.  It also requires users of chef-provisioning to change their cookbooks between versions.  Because this breaks backwards compatibility, according to semver we should only release this change on a major version release.

We could rename all the chef metal resources to `prov_*` (EG, `prov_aws_ebs_volume`) but I don't like that because it requires all cookbook users to rewrite their recipes.  

Instead we could change the class name (`AwsEbsVolume` -> `ProvAwsEbsVolume`) and alias the DSL using the `provides` keyword.  However, this is not backwards compatible to versions of chef older than 11.16ish (I don't remember exactly when the NodeMap was introduced).

@esciara are there any resources the aws cookbook provides that you are trying to use and chef-provisioning does not provide?  If so, another path is to add those resources to chef-provisioning so you do not have a dependency on the aws cookbook.

esciara:
@tyler-ball I am not using the aws cookbook directly. It is called by a cookbook I am using. So I would not be able to say. As to why it calls it, it is a mystery to me!

tyler-ball:
In that case my suggested solution is to use the `provides` syntax.  This would only work with Chef 12 and greater (that is when `provides` was introduced) - would that be a problem?

`lib/chef/provider/aws_ebs_volume.rb`:
```ruby
class Chef::Provider::ProvAwsEbsVolume < Chef::Provider::AwsProvider
  provides :aws_ebs_volume
  ...
```

`lib/chef/resource/aws_ebs_volume.rb`:
```ruby
class Chef::Resource::ProvAwsEbsVolume < Chef::Resource::AwsProvider
  provides :aws_ebs_volume
  ...
```

I'm not sure how the provider resolver would handle this because 2 providers would still map to `aws_ebs_volume` in the recipe DSL.  But it would stop the class name collision :smile:

lynchc:
Eventually we are going to change how we have the resources setup to follow a nested structure.
```
aws.ebs_volume
```
for instance. For now, this simple name change is all that we need. Users don't need to update their cookbooks and, really, this doesn't affect anything from a user point of view.

tyler-ball:
If we go with the current change users would need to update their cookbooks.  If they wanted the aws cookbook resource, they would use `aws_ebs_volume` in their cookbook.  If they wanted the chef-provisioning resource, they would use `aws_ebs_vol` in their cookbook.

lynchc:
the class name changed. the resource_name is the same. look at the changes

tyler-ball:
Using Chef 12 and the latest chef-provisioning/chef-provisioning-aws (including the change in this PR), I setup the following recipe in a cookbook which does NOT `depends 'aws'`:

```ruby
require 'chef/provisioning/aws_driver'
with_driver 'aws'

aws_ebs_volume "test" do

end
```

produces the following output:

```
NoMethodError
-------------
No resource or method named `aws_ebs_volume' for `Chef::Recipe "test"'
```

Using Chef 11.18.0 I get the same error.  If I change my recipe to use `aws_ebs_vol` then it successfully creates the volume.  That is because the `resource_name` attribute in a resource definition does not change the recipe DSL.  The recipe DSL is based entirely off the class name (or the `provides` method in Chef 12).

If you use `aws_ebs_volume` in your recipe and it still works with the change from the PR, that is because it is using the resource from the aws cookbook.

lynchc:
confirmed. learn something new everyday.

@jkeiser, I'll leave this up to you. I don't think using provides is correct either

pburkholder:
I think we need to expect users to keep using the 'aws' cookbook so the node can provision things for itself. So I think we'll see cases where people try to stuff chef-provisioning-aws into a cookbook that already has 'aws' -- which is exactly how I came across it.

Whether they need to both be included in the same chef run is another matter, since the use cases are orthogonal to each other. chef-provisioning-aws is needed on the provisioning node, 'aws' on the already-partially-provisioned node itself.

I don't think any of the solutions above address the problem in a user-friendly way. 

What if we punt here? Catch TypeError here and give a reasonable error message about the 'aws' cookbook before exiting?

 If that becomes too cumbersome, refactor the 'aws' cookbook or adopt the Chef::Resource::AWS:: namespace.

tyler-ball:
The core problem is that there are two classes we are trying to name the same thing, and namespacing seems the obvious answer to that.

The problem with namespacing (Chef::Resource::AWS::* or Chef::Resource::Provisioning::*) is that it breaks the recipe DSL which maps `aws_ebs_volume` to a resource.  Only direct children of Chef::Resource are found by the recipe DSL.  @danielsdeleo / @lamont-granquist let me know if I am wrong about that, but those were my findings with experimentation.

We could do some magic where we add the method `aws` to Chef::Recipe and have that return valid resource instances.  This would create the `aws.ebs_volume` syntax for users.

But I agree with @pburkholder - if we can avoid overlapping these cookbooks, we can avoid this problem for now.

lynchc:
Sounds like everybody is in agreement that we can just close this. Issue #68 can be circumvented by a means of organising cookbooks differently. Correct me if I'm wrong

tyler-ball:
:+1:

lamont-granquist:
@tyler-ball the searching for a const off of Chef::Resource is the fall-back now.   If you use a provides line in the resource you can throw it anywhere in the class heirarchy.

jkeiser:
Question: what if we modified the aws cookbook so it just used chef-provisioning, and made sure we supported the same attributes?

tyler-ball:
I think the better answer is to deprecate the `aws` cookbook and have people start using `chef-provisioning` instead.  They both solve the same problem - using Chef to manage AWS instances/resources.  But if we tried to update the `aws` cookbook to wrap `chef-provisioning` we are still going to run into this namespacing issue, unless we tried some shenanigans.  

One feature we could add to `chef-provisioning-aws` is the ability to read AWS creds out of a databag, like the `aws` cookbook features.

robcoward:
@tyler-ball I don't see that as an option unless I've missed something about chef-provisioning-aws. Our use-case for the aws cookbook is that for EC2 instances we provision using CloudFormations (ie not chef-provisioning) we use it to enforce setting up of tags on the instances. In addition, we make use of the aws_s3_file LWRP to retrieve files from S3. How would those requirements be catered for by chef-provisioning-aws when I'm not looking for the instances to be provisioned by chef-provisioning ?

tyler-ball:
`chef-provisioning` should be able to add tags to previously created instances, if it was able to find the instance correctly.  It may not be able to do this now (because it depends on data bags to find existing instances) but it should be able to do it in the future.  I can't give an ETA on that though.

The non-provisioning resources in the `aws` cookbook (like aws_s3_file) could stay around, and the `aws` cookbook is the right place for them.  So the `aws` cookbook should stay around.

But any AWS resources that can be provisioned or managed via the cookbook should be capable of being managed from chef-provisioning.  If they cannot, it is a bug against chef-provisioning and we should work to fix it.

PR #72, opened 2015-01-28
Ensure we pass a Hash to aws-sdk instances

schisamo:
Classes in the aws-sdk library make the assumption that options are a plain old Hash instance as opposed to an instance of `Cheffish::MergedConfig`. In particular `#delete` is called by many `AWS::EC2::*` classes.

This appears to fix the following issue: https://gist.github.com/schisamo/c6a86e6b0369bb9ef778

/cc @jkeiser @afiune

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -213,7 +213,7 @@ module AWSDriver
           image_options[:instance_id] ||= machine_spec.location['instance_id']
           image_options[:description] ||= "Image #{image_spec.name} created from machine #{machine_spec.name}"
           Chef::Log.debug "AWS Image options: #{image_options.inspect}"
-          image = ec2.images.create(image_options)
+          image = ec2.images.create(image_options.to_hash)
           image_spec.location = {
             'driver_url' => driver_url,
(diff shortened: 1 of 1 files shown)


tyler-ball:
This code looks :+1: to me, but I'm worried about uncovering this error again in the future.  Perhaps we could define a `delete` method on the class in question and have it throw a better error?  Like `Delete is not a valid method, call .to_hash to make a writeable copy` or something?

schisamo:
@tyler-ball perhaps. I think it was a bug we were passing an instance of `Cheffish::MergedConfig` to a method that expected a `Hash`. It might be even better to have `Cheffish::MergedConfig` just extend `Hash`.

tyler-ball:
Ah, I assumed `MergedConfig` didn't want to implement `delete` at all!  Or at least we didn't want AWS deleting items out of it.

In either case, I'm still fine with this PR - we could file an improvement for extending `Hash` if @jkeiser thinks that is wise.

jkeiser:
Yeah.  MergedConfig needs to be replaced with a better thing; I have an OverrideableHash class that could be that thing.

jkeiser:
(i.e. just extend Hash.)

PR #80, opened 2015-02-02
Make machine_batch detect image_id like machine does

jkeiser:
This fails:

```ruby
machine_batch 'blah' do
  machine 'blah' do
  end
end
```

This succeeds:

```ruby
machine 'blah' do
end
```

Both should succeed.

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -61,6 +61,7 @@ module AWSDriver
 
     # Load balancer methods
     def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
+      lb_options ||= {}
       if lb_options[:security_group_id]
         security_group = ec2.security_groups[:security_group_id]
       elsif lb_options[:security_group_name]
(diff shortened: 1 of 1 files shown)


christinedraper:
Will this fix #89?

jkeiser:
@christinedraper nope, I don't think so.  But that's merged now :)

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -61,6 +61,7 @@ def self.canonicalize_url(driver_url, config)
 
     # Load balancer methods
     def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
+      lb_options ||= {}
```
tyler-ball: I prefer to see `lb_options = {}` in the method signature, but thats pretty personal.

tyler-ball:
Other than minor style comment, :+1:

lib/chef/provisioning/aws_driver/driver.rb
jkeiser: Hmm, depends if machine_specs can ever not be passed.  I'll take a look at the caller and see if they can simply be persuaded to pass `{}` themselves.

jkeiser: Note: I have submitted a PR to make it always pass machine_specs.

PR #87, opened 2015-02-06
Fix so image id is taken from bootstrap options

christinedraper:
Fix following problem by using :image_id from bootstrap_options in allocate_machine, so that the following works (ie instantiates appserver from the created image).

machine_image "appserver_image"
end

machine "appserver"
   from_file "appserver_image"
end


christinedraper:
Oops. Duplicate of #47, closing

christinedraper:
Reopening in light of issue #89

jkeiser:
This is surely safer.  I'll file a bug to deprecate the machine_options version.  :+1:

jkeiser:
Rebased and pushed.

PR #90, opened 2015-02-08
Support load_balancers and desired_capacity in aws_auto_scaling_group

christinedraper:
Added load_balancers attribute, as this is a mainstream scenario for autoscaling. Also fixed so desired_capacity attribute is actually passed through to ec2, and updated a converge message to use the group name rather than undefined id.


jkeiser:
:+1:

jkeiser:
Merged (rebased first, so it didn't automatically close).  This is awesome, thanks!

PR #93, opened 2015-02-09
Initial implementation of winrm support

erikvanbrakel:
Porting Windows Remote Management support from https://github.com/chef/chef-provisioning-fog.

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -2,8 +2,10 @@ require 'chef/mixin/shell_out'
 require 'chef/provisioning/driver'
 require 'chef/provisioning/convergence_strategy/install_cached'
 require 'chef/provisioning/convergence_strategy/install_sh'
+require 'chef/provisioning/convergence_strategy/install_msi'
 require 'chef/provisioning/convergence_strategy/no_converge'
 require 'chef/provisioning/transport/ssh'
+require 'chef/provisioning/transport/winrm'
 require 'chef/provisioning/machine/windows_machine'
 require 'chef/provisioning/machine/unix_machine'
 require 'chef/provisioning/machine_spec'
@@ -263,6 +265,26 @@ module AWSDriver
       end
     end
 
+    def user_data
+      # TODO: Make this use HTTPS at some point.
+      <<EOD
+<powershell>
+winrm quickconfig -q
+winrm set winrm/config/winrs '@{MaxMemoryPerShellMB="300"}'
+winrm set winrm/config '@{MaxTimeoutms="1800000"}'
+winrm set winrm/config/service '@{AllowUnencrypted="true"}'
+winrm set winrm/config/service/auth '@{Basic="true"}'
+
+netsh advfirewall firewall add rule name="WinRM 5985" protocol=TCP dir=in localport=5985 action=allow
+netsh advfirewall firewall add rule name="WinRM 5986" protocol=TCP dir=in localport=5986 action=allow
+
+net stop winrm
+sc config winrm start=auto
+net start winrm
+</powershell>
+EOD
+    end
+
     # Machine methods
     def allocate_machine(action_handler, machine_spec, machine_options)
       actual_instance = instance_for(machine_spec)
@@ -274,11 +296,23 @@ module AWSDriver
           Chef::Log.debug('No key specified, generating a default one...')
           bootstrap_options[:key_name] = default_aws_keypair(action_handler, machine_spec)
         end
+
+        puts "#{machine_options.inspect}"
+
+        if machine_options[:is_windows]
+          Chef::Log.info "Setting winRM userdata..."
+          bootstrap_options[:user_data] = user_data
+        else
+          Chef::Log.info "Non-windows, not setting userdata"
+        end
+
         Chef::Log.debug "AWS Bootstrap options: #{bootstrap_options.inspect}"
 
         action_handler.perform_action "Create #{machine_spec.name} with AMI #{image_id} in #{@region}" do
           Chef::Log.debug "Creating instance with bootstrap options #{bootstrap_options}"
+
           instance = ec2.instances.create(bootstrap_options.to_hash)
+
           # Make sure the instance is ready to be tagged
           sleep 5 while instance.status == :pending
           # TODO add other tags identifying user / node url (same as fog)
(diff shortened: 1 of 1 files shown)


lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -274,11 +296,23 @@ def allocate_machine(action_handler, machine_spec, machine_options)
           Chef::Log.debug('No key specified, generating a default one...')
           bootstrap_options[:key_name] = default_aws_keypair(action_handler, machine_spec)
         end
+
+        puts "#{machine_options.inspect}"
```
jkeiser: Need to get rid of putses before we go in :)

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -498,6 +535,53 @@ def default_ami_for_region(region)
       end
     end
 
+    def create_winrm_transport(machine_spec, machine_options, instance)
+      remote_host = determine_remote_host(machine_spec, instance)
+      puts "remote host: #{remote_host}"
```
jkeiser: ^

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -274,11 +296,23 @@ def allocate_machine(action_handler, machine_spec, machine_options)
...
           bootstrap_options[:key_name] = default_aws_keypair(action_handler, machine_spec)
         end
+
+        puts "#{machine_options.inspect}"
+
+        if machine_options[:is_windows]
+          Chef::Log.info "Setting winRM userdata..."
+          bootstrap_options[:user_data] = user_data
+        else
+          Chef::Log.info "Non-windows, not setting userdata"
```
jkeiser: Can you make these two `.debug` statements?  They don't seem like actionable information *most* of the time.

tyler-ball:
This LGTM - did you do some manual testing with it?

erikvanbrakel:
@tyler-ball Yep. Needed windows support as I need to spin-up a bunch of IIS servers, actually provisioned some already. Running into a segfault in chef-client right now, but apparently that's an older bug which has been fixed in chef-client 12.1.0 (https://github.com/chef/ohai/issues/445). Also running into some minor things with load balancers, working on that as well.
I got the CCLA signed and ready to go as well, but that's not linked to this project yet I guess?

jkeiser:
Two things that would be super awesome are autodetection of is_windows, and encrypted winrm with some sort of shared certificate.  But don't worry about that right now, obviously :)

:+1: once you have the puts's removed, assuming it's working for you :)  Nice job!

lib/chef/provisioning/aws_driver/driver.rb
erikvanbrakel: Yeah, oops, will go through it again. I'm fresh into the ruby world, is there an easier way to debug this other than old-school print-to-console logging?

jkeiser:
Actually, I'll go ahead and remove them when I commit.

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: I enjoy using the [pry](https://github.com/pry/pry) gem for debugging.  To use this with projects, I add
```ruby
gem "pry"
gem "pry-byebug"
gem "pry-stack_explorer"
```
to the Gemfile.  Then instead of your puts line above you can use `require 'pry'; binding.pry`.  This will open up an interactive pry session (similar to IRB) which allows you to inspect the current program state.  The 2 other gems also add step through and breakpoint functionality.  I find it quite robust!

jkeiser:
Or that!  Awesome :)

erikvanbrakel:
Not so sure about that userdata powershell though. Don't really see why two ports need to be opened. But I might be missing something. This is how it is in chef-provisioning-fog anyhow...

erikvanbrakel:
Also, I didn't test if ssh still works... Still working on this ;-)

jkeiser:
It is pushed!  (I rebased it first so it wasn't automatically closed.)  Please let us know how the rest of your testing goes :)  The user_data is the same as before, and we passed it over a few more eyeballs and it looks right.  I can't remember the reason for two ports either.

smurawski:
@erikvanbrakel Two ports are being opened in case SSL is used for the transport.

erikvanbrakel:
@smurawski Thanks for the clarification :-)

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -263,6 +265,26 @@ def destroy_image(action_handler, image_spec, image_options)
...
 
+    def user_data
+      # TODO: Make this use HTTPS at some point.
+      <<EOD
+<powershell>
+winrm quickconfig -q
+winrm set winrm/config/winrs '@{MaxMemoryPerShellMB="300"}'
+winrm set winrm/config '@{MaxTimeoutms="1800000"}'
+winrm set winrm/config/service '@{AllowUnencrypted="true"}'
+winrm set winrm/config/service/auth '@{Basic="true"}'
```
btm: I just want to be sure we're not ended up setting up instances in a way that's incompatible with knife-windows/winrm-s as part of our efforts there to standardize on more secure defaults. A while ago I ran into unexpected errors with these settings against development versions of those tools and I need to circle around on their state now.

https://github.com/chef/winrm-s/issues/13

adamedx: Yeah, btw we should try not to use allowenencrypted + basic -- we should eventually (maybe not in this pr) use negotiate instead. If you use SSL, we can use negotiate rather than basic. See the auth table at https://github.com/chef/knife-windows/blob/master/README.md. Basically, allowunencrypted  + basic is fine for test-kitchen, but not for provisioning real infra.To configure SSL, you'll need to create a cert on the system running provisioning and then send it to the remote system and configure a listener.
MaxMemoryPerShellMB should be a lot higher -- in fact I think the default is 1G on Server 2012 and later. I'm assuming

erikvanbrakel:
@adamedx agreed. It works for me for now as I am only provisioning some (non-public) test environments, but once I get to the public facing servers I would like secure communication.

I don't know about the MaxMemoryPerShellMB, simply took it from the chef-provisioning-fog implementation. Up for tweaking I'd say ;-)

PR #96, opened 2015-02-10
Add scheme and subnets to loadbalancer

tfsbuild:
The subnets option allows for creating a loadbalancer in a different subnet, to put it in another VPC than the default one.
The scheme option allows for creating an internal loadbalancer, as opposed to the default which is internet-facing (even though the docs say otherwise)

ref: http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/ELB/LoadBalancerCollection.html#create-instance_method


tfsbuild:
Wrong user.

PR #97, opened 2015-02-10
Add scheme and subnets to loadbalancer

erikvanbrakel:
The subnets option allows for creating a loadbalancer in a different subnet, to put it in another VPC than the default one.
The scheme option allows for creating an internal loadbalancer, as opposed to the default which is internet-facing (even though the docs say otherwise)

ref: http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/ELB/LoadBalancerCollection.html#create-instance_method

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -65,7 +65,7 @@ module AWSDriver
       if lb_options[:security_group_id]
         security_group = ec2.security_groups[:security_group_id]
       elsif lb_options[:security_group_name]
-        security_group = ec2.security_groups.filter('group-name', lb_options[:security_group_name])
+        security_group = ec2.security_groups.filter('group-name', lb_options[:security_group_name]).first
       end
 
       availability_zones = lb_options[:availability_zones]
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -70,11 +70,15 @@ module AWSDriver
 
       availability_zones = lb_options[:availability_zones]
       listeners = lb_options[:listeners]
+      subnets = lb_options[:subnets]
+      scheme = lb_options[:scheme]
 
       lb_optionals = {}
       lb_optionals[:security_groups] = [security_group] if security_group
       lb_optionals[:availability_zones] = availability_zones if availability_zones
       lb_optionals[:listeners] = listeners if listeners
+      lb_optionals[:subnets] = subnets if subnets
+      lb_optionals[:scheme] = scheme if scheme
 
       actual_elb = load_balancer_for(lb_spec)
       if !actual_elb.exists?
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -63,9 +63,9 @@ module AWSDriver
     def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
       lb_options ||= {}
       if lb_options[:security_group_id]
-        security_group = ec2.security_groups[:security_group_id]
+        security_groups = ec2.security_groups[:security_group_id]
       elsif lb_options[:security_group_name]
-        security_group = ec2.security_groups.filter('group-name', lb_options[:security_group_name]).first
+        security_groups = ec2.security_groups.filter('group-name', lb_options[:security_group_name])
       end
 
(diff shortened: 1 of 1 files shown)


erikvanbrakel:
Also wondering... I want to be able to enable/disable sticky sessions using the resource as well. As far as I can tell I have to do this with the AWS ELB client. What would be the preferred syntax for that?

ref: http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/ELB/Client.html#create_app_cookie_stickiness_policy-instance_method

erikvanbrakel:
Diving a bit deeper, the stickiness is actually stored as a load-balancer policy. This also includes SSLNegotiation, ProxyProtocol and some other things. Should it be a collection of policies, or should there be an abstraction on top of that...

erikvanbrakel:
Something like this could work:

    load_balancer "My-LoadBalancer" do
        load_balancer_options :security_group_name => 'webserver',
                              :subnets => 'subnet-01234567',
                              :scheme => 'internal',
                              :policies => [
                                    {
                                        :name =>'cookieStickiness',
                                        :policy_type_name => 'LBCookieStickinessPolicyType',
                                        :policy_attributes => [{ :attribute_name => 'CookieExpirationPeriod', :attribute_value => '10'}]
                                    }
                              ],
                              :listeners => [{
                                    :port => 80,
                                    :protocol => :http,
                                    :instance_port => 80,
                                    :instance_protocol => :http
                                }]
        machines ['a','b']
    end

As a matter of fact, I have something which works with this, sort of. One thing I am not sure about is if the policies should be there, or in the listener. Apparently there's a distinction within these policies. There's methods for enabling a policy for the back end server, or for the listener. To me it seems that this depends on using VPCs or Classic.

So the other option would be:

    load_balancer "My-LoadBalancer" do
        load_balancer_options :security_group_name => 'webserver',
                              :subnets => 'subnet-01234567',
                              :scheme => 'internal',
                              :listeners => [{
                                    :port => 80,
                                    :protocol => :http,
                                    :instance_port => 80,
                                    :instance_protocol => :http,
                                    :policies => [
                                          {
                                              :name =>'cookieStickiness',
                                              :policy_type_name => 'LBCookieStickinessPolicyType',
                                              :policy_attributes => [{ :attribute_name => 'CookieExpirationPeriod', :attribute_value => '10'}]
                                          }
                                    ],
                                }]
        machines ['a','b']
    end

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -64,16 +64,20 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
       if lb_options[:security_group_id]
         security_group = ec2.security_groups[:security_group_id]
       elsif lb_options[:security_group_name]
-        security_group = ec2.security_groups.filter('group-name', lb_options[:security_group_name])
+        security_group = ec2.security_groups.filter('group-name', lb_options[:security_group_name]).first
```
jkeiser: Is there an issue if this returns multiple groups?  It looks like the later code expects an array anyway ...

jkeiser:
Just the one question, and otherwise looks great.

In response to the policies thing ... can we get away with not requiring explicit values and passing a hash through (like for lb_optionals)?  Then policies and other things could just be "whatever the load balancer supports" and we wouldn't have to add code every time there was something new ...

erikvanbrakel:
I'll take a look at the policies stuff this weekend. Sounds like a good idea to just pass in the hash of values that will be passed to the AWS SDK, but it would probably be an array. Multiple policies can be applied, but as far as I know you have to add (and enable) them one by one.

lib/chef/provisioning/aws_driver/driver.rb
erikvanbrakel: Ah, I think I wrongly diagnosed the bug. The bug is not so much in the calling to the AWS API, but one of the log writes assumes a singular result. Thus this should work with multiple security groups, but fails. Will fix, brb.

erikvanbrakel:
Alright, made some changes there. It now works if I pass a single or multiple sec. groups, although in my case the AWS API complains about limits. Might be a related but different bug...

erikvanbrakel:
Probably a better idea to take the policy stuff to another PR, so this can be merged

jkeiser:
@erikvanbrakel looking good!  Merging :)

tyler-ball:
Closing this as it will be merged as part of https://github.com/chef/chef-provisioning-aws/pull/108

PR #98, opened 2015-02-11
aws_ebs_volume enhancements

patrick-wright:
I picked up where this LWRP left off. 
* Updated :create to support more ebs config options
* Added :attach and :detach actions
* Added some testing recipes to ensure basic functionality
* Verified volumes can be created / attached and detached / deleted within same resource


lib/chef/provider/aws_ec2_volume.rb
```diff
@@ -0,0 +1,137 @@
+require 'chef/provider/aws_provider'
+require 'cheffish'
+require 'date'
+require 'retryable'
+
+class Chef::Provider::AwsEc2Volume < Chef::Provider::AwsProvider
```
patrick-wright: Breaking change, but this resource was limited in functionality to begin with.

lib/chef/provider/aws_ec2_volume.rb
```diff
@@ -0,0 +1,137 @@
...
+          )
+          new_resource.attached_to_instance new_resource.instance_id
+          new_resource.attached_to_device new_resource.device
+
+          wait_for_volume_status(:in_use)
+        end
+      rescue AWS::EC2::Errors::VolumeInUse => e
+        # assume attached correctly, still need to check if its the intended instance
+        Chef::Log.debug(e.message)
+        new_resource.updated_by_last_action(true)
```
patrick-wright: remove line

lib/chef/provider/aws_ec2_volume.rb
```diff
@@ -0,0 +1,137 @@
...
+            # todo load current resource and remove keys
+            new_resource.attached_to_instance ''
+            new_resource.attached_to_device ''
+
+            wait_for_volume_status(:available)
+        end
+      rescue AWS::EC2::Errors::IncorrectState => e
+        # assume detached correctly
+        Chef::Log.debug(e.message)
+        new_resource.updated_by_last_action(true)
```
patrick-wright: remove line

patrick-wright:
I really should check for volume state, and correct instance ids before calling api to allow exceptions to be caught.

spec/acceptance/aws_ec2_volume/cookbooks/aws_ec2_volume_fixture/recipes/gambit.rb
```diff
@@ -0,0 +1,47 @@
+include_recipe 'aws_ec2_volume_fixture'
+
+ebs_test_node = search(:node, "*:#{node['test']}").first
+
+volume1 = aws_ec2_volume node['test'] do
+  action [:create, :attach, :detach, :delete]
```
patrick-wright: change to :nothing

spec/acceptance/aws_ec2_volume/cookbooks/aws_ec2_volume_fixture/recipes/gambit.rb
```diff
@@ -0,0 +1,47 @@
...
+  availability_zone 'us-west-2a'
+  size 10
+  volume_type :io1
+  iops 300
+  instance_id ebs_test_node[:ec2][:instance_id]
+  device '/dev/xvdf'
+end
+
+volume2 = aws_ec2_volume "#{node['test']}-2" do
+  action [:create, :attach, :detach, :delete]
```
patrick-wright: change to :nothing

lib/chef/provider/aws_ec2_volume.rb
jkeiser: Are we doing this to avoid conflict with the aws cookbook?

jkeiser:
I am updating this to aws_ebs_volume for the moment just so people don't get in the habit of typing the new pattern: if we switch things to avoid conflict with the AWS cookbook, we should do it all at once.

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -27,14 +37,71 @@ class Chef::Provider::AwsEbsVolume < Chef::Provider::AwsProvider
 
   action :delete do
     if existing_volume
-      converge_by "Deleting EBS volume #{fqn} in #{new_resource.region_name}" do
-        existing_volume.delete
+      begin
```
jkeiser: What is `begin` here for?  I don't see an ensure / rescue ...

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -27,14 +37,71 @@ class Chef::Provider::AwsEbsVolume < Chef::Provider::AwsProvider
...
+      converge_by "Attaching EBS volume #{fqn} in #{new_resource.region_name} to instance #{new_resource.instance_id}" do
+
+        existing_volume.attach_to(
+          ec2.instances[new_resource.instance_id],
+          new_resource.device
+        )
+        new_resource.attached_to_instance new_resource.instance_id
+        new_resource.attached_to_device new_resource.device
+
+        wait_for_volume_status(:in_use)
```
jkeiser: We are not checking for the right thing here: we want to wait for the volume to be attached to this particular instance, yes?  What happens if you call attach and then someone else calls detach/attach before you check for in use?

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -53,7 +120,25 @@ def existing_volume
   end
 
   def id
-    new_resource.volume_name
+    new_resource.name
+  end
+
+  def volume_attached_to_configured_instance?
+    detected = @existing_volume.attachments.detect { |attachment|
+      attachment.instance.id == new_resource.instance_id
```
jkeiser: What if the instance is the same but the device is different?

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -29,12 +39,67 @@ class Chef::Provider::AwsEbsVolume < Chef::Provider::AwsProvider
...
+          raise TimeoutError,
+            "Timed out waiting for volume #{new_resource.volume_id} to be deleted." if existing_volume.exists?
+        end
       end
     end
 
     new_resource.delete
   end
 
+  action :attach do
```
jkeiser: Typically, the way we handle this is more descriptive: we'd handle the attach as part of the `create`:

```ruby
aws_ebs_volume 'volume' do
  ...
  attached_to_instance 'i-234234234'
end
```

And the create action would both create and attach the instance.  No need for a separate `attach` action.

Also, we should let you attach to a `machine` instead of an `instance`.

patrick-wright: @jkeiser, Cool. I see wanting to attach to an instance upon creation. I also see your point about `machine`.
How about:
```ruby
aws_ebs_volume 'volume' do
  machine 'blah'
end

aws_ebs_volume 'another_volume' do
  instance 'i-12345678'
end

```

lib/chef/provider/aws_ebs_volume.rb
patrick-wright: @jkeiser, not following you on this one.

lib/chef/provider/aws_ebs_volume.rb
patrick-wright: @jkeiser aws-sdk will raise the exception.

lib/chef/provider/aws_ebs_volume.rb
jkeiser: Actually, if you get a chance you might want to look at the jk/no_mo_data_bag_resource branch.  I may have already made machine work there, I did a general sweep through all the resources.

lib/chef/provider/aws_ebs_volume.rb
jkeiser: I was expecting for it to wait until the volume is in use with this particular machine. It's possible for someone else to attach it to a different instance before you get to this line ...

PR #99, opened 2015-02-15
Fix specs require

pburkholder:
With ChefDK 0.4.0 I'm unable to run 'bundle exec rspec spec' until
I add this require.


jkeiser:
The right place to fix is in chef itself (we're not calling ShellOut, so one of the chef files isn't calling require on its own), but I already have a PR out for this one it turns out :)  Merging so we get past the issue.

PR #100, opened 2015-02-16
Specs for provider and resourcs

pburkholder:
I don't know if this is useful, but I thought I would try adding some rspec tests for the providers and resources here, starting with aws_subnet as a random example.

A lot of this is cargo-culting, but I wanted to update some of the providers and add others, and needed a framework to do so safely, and I think this will help.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,15 +2,4 @@ source "https://rubygems.org"
 gemfile
 gemspec
 
-
 gem 'chef-provisioning', :git => 'https://github.com/opscode/chef-provisioning.git', :branch => 'master'
-
-
-
-group :pdb do
-  gem 'nyan-cat-formatter'
-  gem 'pry-nav'
-  gem 'pry-rescue'
-  gem 'pry-stack_explorer'
-end
-
--- a/Gemfile
+++ b/Gemfile
@@ -2,4 +2,15 @@ source "https://rubygems.org"
 gemfile
 gemspec
 
+
 gem 'chef-provisioning', :git => 'https://github.com/opscode/chef-provisioning.git', :branch => 'master'
+
+
+
+group :pdb do
+  gem 'nyan-cat-formatter'
+  gem 'pry-nav'
+  gem 'pry-rescue'
+  gem 'pry-stack_explorer'
+end
+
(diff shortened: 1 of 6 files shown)


erikvanbrakel:
Love this. I've been thinking about this myself, because I really don't want to be running manual tests against AWS just to see if new stuff works.

christinedraper:
Definitely a good idea!

patrick-wright:
@pburkholder Great work!  I've already been implementing your patten in my branches for ec2 volumes and ec2 network interfaces.

pburkholder:
@patrick-wright #101 is better organized spec-wise since I used it to build a new provider.

pburkholder:
@tyler-ball This and #101 are what I've been working on. I wrote this to see if I could write specs around an existing, working resource, and #101 to implement a new resource, internet_gateway, with the tests to help me in the process. 

Regarding data_bags, see https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provider/aws_provider.rb#L25
and
https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/resource/aws_resource.rb#L2

jkeiser:
For what you are doing here, I actually think cheffish might work even better--it lets you test the resource and provider in the context of a recipe: https://github.com/chef/cheffish/blob/master/spec/integration/chef_user_spec.rb#L17-L21

(Note: this is not a C/R requirement, more a suggestion :)

with_converge do
  aws_subnet ... do
  end
end

pburkholder:
@jkeiser I saw that code in Cheffish this weekend, but wasn't sure how to apply in this case, as I don't want to make calls out to AWS to test the code while it's under development. Would that approach work as an integration test while we use mocked API calls, as in this PR, to do unit tests?

Also, what's a "C/R requirement" ?  

Thanks,

Peter

tyler-ball:
I like these tests!  I have the same comment here about `allow_any_instance_of` vs `expect_any_instance_of` that I made in https://github.com/chef/chef-provisioning-aws/pull/101#discussion_r25173814

jkeiser:
@pburkholder oh, "C/R" was just an abbreviation for Code Review.  Just saying the suggestion to switch it to Cheffish certainly won't block a merge, but would probably be a better pattern for the tests to follow.

And yep, the "run the chef client" tests are compatible with mocked API calls :)  The approach of mocking the AWS API is a good one.

tyler-ball:
I like the idea of using the Cheffish spec_helper because it removes a lot of the boilerplate around defining `node`, `run_context`, etc.

jkeiser:
Given that "please try and use Cheffish instead" isn't a blocker, this seems OK to merge, :+1:

jkeiser:
Merged as 48bc0816d80dcdc70c337b3bf77aafa54f23aad8 (I added a "bundle exec rake spec" and .travis.yml as well).

tyler-ball:
:+1:

PR #106, opened 2015-02-24
Bug fixes for creating load balancers

tyler-ball:
I was seeing two bugs when trying to create a load balancer:

1. When I did not use the correct keys for the `listeners` object, I was getting weird errors.  This was caused by https://github.com/chef/chef-provisioning-aws/blob/1a33c43bb81d7a3955d5e3eb657d7c82ce4f9021/lib/chef/provisioning/aws_driver/driver.rb#L130 so I added a validate method to prevent weird errors in the future.
2. `add_listeners` is being populated as a hash from port to listener, but was being accessed as an array - I fixed this with https://github.com/chef/chef-provisioning-aws/blob/7d689836cf0ba1cce4f41301ddcf6a1713c5cdef/lib/chef/provisioning/aws_driver/driver.rb#L160
3. When trying to update the listener object, users can supply either `HTTP` or :http as the protocol.  These should be equivalent so no listener update should be issued.  I fixed this with https://github.com/chef/chef-provisioning-aws/blob/1a33c43bb81d7a3955d5e3eb657d7c82ce4f9021/lib/chef/provisioning/aws_driver/driver.rb#L134
4. Updating the listener on an existing ELB was calling an unknown method `perform_action` - looking at the code, I thought you meant to call this? https://github.com/chef/chef-provisioning-aws/blob/1a33c43bb81d7a3955d5e3eb657d7c82ce4f9021/lib/chef/provisioning/aws_driver/driver.rb#L127
5. listeners are immutable - must delete and recreate if config is modified

@jkeiser I'm still confused by these procs-within-procs to call the actions - can we add some comments or figure out a cleaner way to present this?

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -122,35 +122,37 @@ module AWSDriver
           end
         end
 
-        # Update listeners
-        perform_listener_action = proc do |desc, &block|
-          perform_listener_action = proc { |desc, &block| perform_action.call(desc, &block) }
-        end
+        # Update listeners - THIS IS NOT ATOMIC
         add_listeners = {}
         listeners.each { |l| add_listeners[l[:port]] = l } if listeners
         actual_elb.listeners.each do |listener|
           desired_listener = add_listeners.delete(listener.port)
           if desired_listener
+
+            # listener.(port|protocol|instance_port|instance_protocol) are immutable for the life
+            # of the listener - must create a new one and delete old one
+            immutable_updates = []
             if listener.protocol != desired_listener[:protocol].to_sym.downcase
-              perform_listener_action.call("    update protocol from #{listener.protocol.inspect} to #{desired_listener[:protocol].inspect}'") do
-                listener.protocol = desired_listener[:protocol]
-              end
+              immutable_updates << "    update protocol from #{listener.protocol.inspect} to #{desired_listener[:protocol].inspect}"
             end
             if listener.instance_port != desired_listener[:instance_port]
-              perform_listener_action.call("    update instance port from #{listener.instance_port.inspect} to #{desired_listener[:instance_port].inspect}'") do
-                listener.instance_port = desired_listener[:instance_port]
-              end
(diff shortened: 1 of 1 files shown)


lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -120,42 +122,44 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
           end
         end
 
-        # Update listeners
-        perform_listener_action = proc do |desc, &block|
```
tyler-ball: @jkeiser I didn't see the use in having both this and `perform_action` - what am I missing?

jkeiser: I have no idea, this is bizarre code.  I'd prefer we use perform_action too.

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -120,42 +122,44 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
-                listener.instance_protocol = desired_listener[:instance_protocol]
-              end
+            if listener.instance_protocol != desired_listener[:instance_protocol].to_sym.downcase
+              immutable_updates << "    update instance protocol from #{listener.instance_protocol.inspect} to #{desired_listener[:instance_protocol].inspect}"
             end
-            if listener.server_certificate != desired_listener[:server_certificate]
-              perform_listener_action.call("    update server certificate from #{listener.server_certificate} to #{desired_listener[:server_certificate]}'") do
+            if !immutable_updates.empty?
+              perform_action.call(immutable_updates) do
+                listener.delete
```
jkeiser: I was about to mention idempotency issues, but it seems like it's an Amazon thing in general :)  http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-elb.html#cfn-ec2-elb-listeners

jkeiser:
:+1:

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: Yup, thats the exact same documentation I was looking at that made me give up trying to figure out how to make it atomic :)

PR #108, opened 2015-02-25
Security groups, subnets and schemes can be updated

tyler-ball:
This completes the work started in https://github.com/chef/chef-provisioning-aws/pull/97 - that correctly created a LB with the desired attributes, but did not update an converge an existing LB to the desired state.

\cc @jkeiser @fnichol @erikvanbrakel

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -118,7 +118,7 @@ module AWSDriver
         # TODO: refactor this whole giant method into many smaller method calls
         # Update scheme - scheme is immutable once set, so if it is changing we need to delete the old
         # ELB and create a new one
-        if scheme.downcase != actual_elb.scheme
+        if scheme && scheme.downcase != actual_elb.scheme
           desc = ["  updating scheme to #{scheme}"]
           desc << "  WARN: scheme is immutable, so deleting and re-creating the ELB"
           perform_action.call(desc) do
@@ -151,16 +151,16 @@ module AWSDriver
           enable_zones = (availability_zones || []).dup
           disable_zones = []
           actual_elb.availability_zones.each do |availability_zone|
-            if !enable_zones.delete(availability_zone.name)
+            unless enable_zones.delete(availability_zone.name)
               disable_zones << availability_zone.name
             end
           end
-          if enable_zones.size > 0
+          unless enable_zones.empty?
             perform_action.call("  enable availability zones #{enable_zones.join(', ')}") do
               actual_elb.availability_zones.enable(*enable_zones)
             end
           end
-          if disable_zones.size > 0
+          unless disable_zones.empty?
             perform_action.call("  disable availability zones #{disable_zones.join(', ')}") do
               actual_elb.availability_zones.disable(*disable_zones)
             end
@@ -171,11 +171,11 @@ module AWSDriver
           attach_subnets = (subnets || []).dup
           detach_subnets = []
           actual_elb.subnets.each do |subnet|
-            if !attach_subnets.delete(subnet.id)
+            unless attach_subnets.delete(subnet.id)
               detach_subnets << subnet.id
             end
           end
-          if attach_subnets.size > 0
+          unless attach_subnets.empty?
             perform_action.call("  attach subnets #{attach_subnets.join(', ')}") do
               elb.client.attach_load_balancer_to_subnets(
                 load_balancer_name: actual_elb.name,
@@ -183,7 +183,7 @@ module AWSDriver
               )
             end
           end
-          if detach_subnets.size > 0
+          unless detach_subnets.empty?
             perform_action.call("  detach subnets #{detach_subnets.join(', ')}") do
               elb.client.detach_load_balancer_from_subnets(
                 load_balancer_name: actual_elb.name,
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -69,6 +69,8 @@ module AWSDriver
 
       availability_zones = lb_options[:availability_zones]
       listeners = lb_options[:listeners]
+      subnets = lb_options[:subnets]
+      scheme = lb_options[:scheme]
 
       validate_listeners(listeners)
 
@@ -76,6 +78,8 @@ module AWSDriver
       lb_optionals[:security_groups] = [security_group] if security_group
       lb_optionals[:availability_zones] = availability_zones if availability_zones
       lb_optionals[:listeners] = listeners if listeners
+      lb_optionals[:subnets] = subnets if subnets
+      lb_optionals[:scheme] = scheme if scheme
 
       actual_elb = load_balancer_for(lb_spec)
       if !actual_elb.exists?
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -165,7 +165,7 @@ module AWSDriver
 
           # Only 1 of subnet or AZ will be populated b/c of our check earlier
           desired_subnets_zones = {}
-          (availability_zones || []).each do |zone|
+          availability_zones.each do |zone|
             # If the user specifies availability zone, we find the default subnet for that
             # AZ because this duplicates the create logic
             zone = zone.downcase
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -17,6 +17,7 @@ require 'chef/provisioning/aws_driver/credentials'
 
 require 'yaml'
 require 'aws-sdk-v1'
+require 'set'
 
 # loads the entire aws-sdk
 AWS.eager_autoload!
@@ -130,7 +131,7 @@ module AWSDriver
 
         # Update security groups
         if security_group_ids.empty?
-            Chef::Log.info("No Security Groups specified in resource; load_balancer[#{actual_elb.name}] cannot have " +
+            Chef::Log.debug("No Security Groups specified. Load_balancer[#{actual_elb.name}] cannot have " +
             "empty Security Groups, so assuming it only currently has the default Security Group.  No action taken.")
         else
           current = Set.new(actual_elb.security_group_ids)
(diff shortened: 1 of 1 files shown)


lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -62,28 +62,42 @@ def self.canonicalize_url(driver_url, config)
...
 
-      availability_zones = lb_options[:availability_zones]
+      availability_zones = lb_options[:availability_zones] || []
+      subnets = lb_options[:subnets] || []
       listeners = lb_options[:listeners]
+      scheme = lb_options[:scheme]
+
+      if !availability_zones.empty? && !subnets.empty?
+        raise "You cannot specify both `availability_zones` and `subnets`"
+      end
```
jkeiser: Does the AWS driver already give a reasonable error message for this?  If so, we might want to leave it up to the driver (in general, let the driver own its own logic so it can change the logic itself).

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -101,23 +115,83 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
+        if !availability_zones.empty?
+
+          enable_zones = (availability_zones || []).dup
+          disable_zones = []
+          actual_elb.availability_zones.each do |availability_zone|
+            if !enable_zones.delete(availability_zone.name)
+              disable_zones << availability_zone.name
+            end
+          end
+          if enable_zones.size > 0
```
jkeiser: Can you use empty?  That's what it looks like in the other spots in this method.

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -101,23 +115,83 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
+              actual_elb.availability_zones.enable(*enable_zones)
+            end
           end
+          if disable_zones.size > 0
+            perform_action.call("  disable availability zones #{disable_zones.join(', ')}") do
+              actual_elb.availability_zones.disable(*disable_zones)
+            end
+          end
+
+        elsif !subnets.empty?
```
jkeiser: What happens if the user starts out with a load balancer with subnets, and then changes it to have availability_zones?  (Or vice versa.)  It might be best to unconditionally calculate the enable/disable/attach/detach arrays, so that subnets can be removed and then availability zones added.

jkeiser:
@tyler-ball a few comments and one (maybe?) bug, but generally :+1: :)

jkeiser:
@tyler-ball one more thing, actually: I think the scheme change feature might have unexpected side-effects--you miss tags, and you miss anything that is linked to the load balancer.  Worth a think, but probably still worth having the feature in.

Also, take a look at cloudformation's description of how it does this: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks.html#update-replacement

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: The driver DOES give a good error message on this.  With an older version of my code I was seeing some weird issues switching from availability zones to subnets, but I think I prevent that error now.  Removing this.

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: Thats why I have this check.  If you start with availability zones and switch to subnets, it simply tries to update the subnets.  If you start with subnets and switch to availability zones, it tries to update the availability zones.

If you specify both, it will update first availability zones then subnets - this will possibly cause the resource to converge every time, but I was going to chalk that up to a user error.

If you specify neither then neither the availability zones or subnets change.  I did this because it isn't possible for a load balancer to belong to 0 availability zones (and consequently, it also cannot belong to 0 subnets).

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: Done

tyler-ball:
@jkeiser Sooooo - when changing scheme, my first thought was to clone the original object and then update the scheme.  This would preserve tags but would not preserve instances.  The only way I could find to clone the whole object was http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/ELB/Client.html#describe_load_balancers-instance_method which returns a hash.  This hash would have to manually be transformed into an object - there isn't a `from_json`/`to_h` method I can see in the aws libraries.  There are also fields returned there that are unnecessary - `instance_id`, `created_time`, etc.

I decided to go with my current code because I think if you are going to attach tags/instances to a load balancer, you should be assigning those items inside chef-provisioning.  In this case we would retag the object after creating it with the updated scheme.

Reading about `Replacement` on the AWS side now, I think I should just hold off on deleting the old LB until the new instances are attached to the new LB.  I'll push a change out for that now and get your review.

lib/chef/provisioning/aws_driver/driver.rb
jkeiser: My concern is this: if you start with subnets and switch to availability zones, then it will just add the availability zones, when it *should* delete the subnets as well.  More specifically, I think this case will have an issue:

First converge:
```ruby
load_balancer 'blah' do
  availability_zones [ 'a', 'b', 'c' ]
end
```

Second converge:

```ruby
load_balancer 'blah' do
  subnets [ 'd', 'e', 'f' ]
end
```

I would expect the second recipe to remove the availability zones as well as add the subnets, but I don't think your code will do the removing.

At least, that's how the code reads to me.  I think you want an algorithm more like this:

- Delete any subnets that the user didn't ask for
- Delete any availability zones that the user didn't ask for
- Add any subnets the user asked for (and don't already exist)
- Add any availability zones the user asked for (and don't already exist)

Where the current flow is more like:

- If the user specified availability zones,
  - Delete any availability zones that the user didn't ask for
  - Add any availability zones the user asked for (and don't already exist)
- If the user specified subnets and did NOT specify availability zones,
  - Delete any subnets the user didn't ask for
  - Add any subnets the user asked for (and don't already exist)

Bonus is fewer branches in the logic means more of the logic is part of the normal flow (and therefore tested at least a little bit every time you run).

jkeiser:
I agree on the scheme thing, just wanted to bring it up :)

I still have one comment in there about the subnets vs. availability zones.  I think there may be a bug lurking in there.

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: Ahhhhh, you are right - I understand now.  I'll update the logic to the following flow:

* Add any subnets the user asked for (and don't already exist)
* Add any availability zones the user asked for (and don't already exist)
* Delete any subnets that the user didn't ask for
* Delete any availability zones that the user didn't ask for

If you delete before adding and you remove all subnets or availability zones, AWS throws an error. 

Again, if they specify both subnets and availability zones (even if adding to one and removing from another) there is a possibility they may overlap.  EG, they may add availability zone `b` and remove subnet `123`, which removes the ELB from availability zone `b`.

lib/chef/provisioning/aws_driver/driver.rb
jkeiser: Oh heck, that forced you to move to unless.  But I'm going to stop thrashing this C/R with nits.

tyler-ball:
Okay, I reworked the availabilty_zone/subnet stuff again.  My logic is this - because subnets belong to an availability zone, I create 2 maps.  The keys are availability zones and the values are the set of subnets that reside in that availability zone.  One map is for current state and one map is for desired state.  A desired state that has a key (availability zone) where the value is the empty set means we want that availability zone and the default subnet there.

Then I diff these 2 maps - items in the desired but not the current are added, then items in the current but not in the desired are removed.

Sound good?  This should be ready for a final review.

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -101,22 +113,128 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
+            if desired_zones_subnets.keys.include?(zone)
+              # Don't remove the subnet if it would remove a desired zone
+              unless desired_zones_subnets[zone].empty?
+                detach_subnets += (subnets - desired_zones_subnets[zone]).to_a
+              end
+            else
+              disable_zones << zone unless availability_zones.empty?
+              detach_subnets += subnets.to_a
+            end
+          end
```
tyler-ball: Lines 213-223 of this show my removal logic - this is where I made some assumptions.

1. If the past state specified 1 or more subnets in availability zone A, and the new state only specifies availability zone A (IE, switching from managing subnets to managing zones) then I don't remove any of the subnets.  I do this because there is no way to determine what the default subnet is from a given availability zone.  `subnets: ['123', '456']` -> `availability_zones: ['a']` = no change if subnets `123` and `456` are in zone A.
2. Similarly, if they move from `availability_zones: ['a']` -> `subnets: ['123', '456']` where `123` and `456` live in Zone A, I only add the missing subnets.
  1. However, if `123` is the default subnet and they move `availability_zones: ['a']` -> `subnets: ['456']` I *will* remove subnet `123`.  I do not touch the availability zone in this instance.
3. If they move from managing availability zones to managing subnets, I will only send add/remove subnet calls.  EG,  `availability_zones: ['a']` -> `subnets: ['789']` will send a 'remove subnet 123' and 'add subnet 789' call to AWS.  It will not send any add/remove calls for availability zones.
  1. However!  If they move from
```ruby
availability_zones: ['a']
```
to
```ruby
availability_zones: ['b']
subnets: ['789']
```
I will send AWS 'remove availability zone a', 'remove subnet 123', 'add availability zone b', 'add subnet 789'

Let me know if you don't agree with this logic.

jkeiser:
Yeah, that algorithm sounds like what I normally do.

lib/chef/provisioning/aws_driver/driver.rb
jkeiser: I don't understand #1:

Scenario A:
```ruby
load_balancer 'blah' do
  subnets '456', '789' # both in availability zone A, 
end

load_balancer 'blah' do
  availability_zones 'A'
end
```

Scenario B:

```ruby
load_balancer 'blah' do
  availability_zones 'A'
end
```

If the default subnet for A is 123, then these two recipes produce different results, even though the resource is the same in both.

tyler-ball:
Okay, I made ELB update work the same way as create and it simplified the code a lot.  I also rebased against master.  This is ready for review!

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -102,22 +117,111 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
           end
         end
-        if enable_zones.size > 0
-          perform_action.call("  enable availability zones #{enable_zones.join(', ')}") do
-            actual_elb.availability_zones.enable(*enable_zones)
+
+        # Update security groups
+        if security_group_ids.empty?
+            Chef::Log.debug("No Security Groups specified. Load_balancer[#{actual_elb.name}] cannot have " +
+            "empty Security Groups, so assuming it only currently has the default Security Group.  No action taken.")
```
jkeiser: I think this will be wrong in the case of update: if the ELB has security groups other than default, then we should remove them, shouldn't we?  i.e. what if they added security groups but decided they wanted to go back to default ...

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -102,22 +117,111 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
+
+        # A subnet always belongs to an availability zone.  When specifying a ELB spec, you can either
+        # specify subnets OR AZs but not both.  You cannot specify multiple subnets in the same AZ.
+        # You must specify at least 1 subnet or AZ.  On an update you cannot remove all subnets
+        # or AZs - it must belong to one.
+        if !availability_zones.empty? && !subnets.empty?
+          # We do this check here because there is no atomic call we can make to specify both
+          # subnets and AZs at the same time
+          raise "You cannot specify both `availability_zones` and `subnets`"
+        else
```
jkeiser: You don't need this else-after-raise, and it may help the code to reduce indentation :)

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -102,22 +117,111 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
+            action = "  attach subnets #{attach_subnets.join(', ')}"
+            enable_zones = (desired_subnets_zones.map {|s,z| z if attach_subnets.include?(s)}).compact
+            action += " (availability zones #{enable_zones.join(', ')})"
+            perform_action.call(action) do
+              begin
+                elb.client.attach_load_balancer_to_subnets(
+                  load_balancer_name: actual_elb.name,
+                  subnets: attach_subnets
+                )
+              rescue AWS::ELB::Errors::InvalidConfigurationRequest
```
jkeiser: Awesome error message!  We can handle this by re-running and doing the create-a-new-lb dance, can't we?  Again, new issue, doesn't block merge.  I like that you are exploring the update side of things; this work is going to help us figure out how to update things in general.

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -102,22 +117,111 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
+          unless subnets.empty?
+            subnet_query = ec2.client.describe_subnets(:subnet_ids => subnets)[:subnet_set]
+            # AWS raises an error on an unknown subnet, but not an unknown AZ
+            subnet_query.each do |subnet|
+              zone = subnet[:availability_zone].downcase
+              desired_subnets_zones[subnet[:subnet_id]] = zone
+            end
+          end
+
+          # We only bother attaching subnets, because doing this automatically attaches the AZ
```
jkeiser: Oh wow!  So what you are saying is, when you attach the subnet AZ-1-defaultsubnet, and then describe the LB, amazon reports it as attached to the AZ itself rather than the subnet?

jkeiser:
Many comments, but except for one trivial change and a question, nothing that would get in the way of merge :)

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: You're right - this is also different than the create case.  I wrote this before we went down the AZ/subnet rabbit hole and fell into aws-everland

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: Too true!

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: Correct, the way around this is to create a new LB and move instances into that LB.  For now the users would have to do that manually, but the goal is for us to eventually replicate cloudformation and do it for them.

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: After attaching the subnet AZ-1-defaultsubnet and describing the LB, you can see the AZ is AZ-1 and the subnet is AZ-1-defaultsubnet.  The describe returns both, but attaching to the subnet also attaches to the AZ.  And vice-versa.

tyler-ball:
Okay, subnets now reset to the default if the attribute is removed from the resource.  If this looks good, go ahead and hit the merge button!

tyler-ball:
I am now confused - why are we resetting `security_group` back to the default if the user doesn't supply it?  We don't do that with other resources.  If I create a `directory` resource it will get the default linux permissions (lets assume 755).  If I then update the mode attribute to '700' and run chef, it will apply the change.  If I then remove that attribute from the resource, it doesn't reset it to the default of '755'.

So why should we reset security groups to the default if the attribute is missing?

jkeiser:
Yeah, that seems reasonable.  If you don't specify an attribute, then we shouldn't make a change.

lib/chef/provisioning/aws_driver/driver.rb
jkeiser: Oh!  Does that happen when you attach to a non-default subnet too?

jkeiser:
@tyler-ball we should be clear, though: if the user specifies `security_groups []`, that is a different case from not specifying the `security_groups` at all.  In the former case, removing all security groups, erroring, or maybe reattaching the default (dunno what is right) is what they asked for; in the latter, they asked for the security groups to remain as they are.

jkeiser:
Let's keep talking about it, but I'll merge this as "waaay better than it was."  Easier to think about the smaller things when we've got this there.

jkeiser:
Or not ... my tree is rather full at the moment.  Feel free to merge when you get a moment; I'll chat you in the morning :)

Issue #109, opened 2015-02-25
Exceptions when try to update load balancer


christinedraper:
Using latest in master, I have a load balancer:

```
load_balancer "elbappserverelb" do
  load_balancer_options({
  :availability_zones=>["us-west-2c", "us-west-2b", "us-west-2a"], 
  :listeners=>[{:instance_port=>3001, :protocol=>"HTTP", :instance_protocol=>"HTTP", :port=>3001}]})
  machines([])
end
```

Load balancer was created, but on rerun, I got an exception:
```
NoMethodError: load_balancer[elbappserverelb] (@recipe_files::/home/christine/provisioning/cfgen/aicf.rb line 7) had an error: NoMethodError: undefined method `perform_action' for #<Chef::Provisioning::AWSDriver::Driver:0x00000007c099d0>
/home/christine/.chefdk/gem/ruby/2.1.0/gems/chef-provisioning-aws-0.2.2/lib/chef/provisioning/aws_driver/driver.rb:125:in `block (2 levels) in allocate_load_balancer'
/home/christine/.chefdk/gem/ruby/2.1.0/gems/chef-provisioning-aws-0.2.2/lib/chef/provisioning/aws_driver/driver.rb:143:in `call'
/home/christine/.chefdk/gem/ruby/2.1.0/gems/chef-provisioning-aws-0.2.2/lib/chef/provisioning/aws_driver/driver.rb:143:in `block in allocate_load_balancer'
```
Not sure how to fix the above, sorry.

I also got an exception thrown on L158 

https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/driver.rb#L158

Which I believe needs to be:

add_listeners.each do |port, listener|

christinedraper:
Figured it out - I'll submit a PR

PR #110, opened 2015-02-25
fix two exceptions in driver on lb update

christinedraper:


christinedraper:
To address issue #109

PR #111, opened 2015-02-27
Add `driver` and `with_driver` support to AWS resources

### LINKED ISSUE #75: AWS Profile not honored for aws_vpc (and possibly others)
I created an AWS .credentials file with two profiles, one has authority to create VPCs, one didn't. 

I tried various ways to set the profile, including with_driver 'aws:test1' and setting CHEF_DRIVER. In all cases, it ignored the specified profile and used the default one.

The same driver URL worked to set the profile for the machine resource.

When I set a breakpoint in AwsDriver.initialize, it appear not even to be called in the case of aws_vpc, although it is for machine.

I am using chef-provisioning-aws 0.2.1 and chef-provisioning 0.17

jkeiser:
Remove `region` as a resource attribute and remove `with_data_center`
Fixes #75

### DIFF
--- a/docs/examples/auto_scaling.rb
+++ b/docs/examples/auto_scaling.rb
@@ -1,7 +1,6 @@
 require 'chef/provisioning/aws_driver'
-with_driver 'aws'
 
-with_data_center 'eu-west-1' do
+with_driver 'aws::eu-west-1' do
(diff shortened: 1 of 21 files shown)


fnichol:
That's looking much more cleaned up. :+1:

PR #112, opened 2015-02-27
Support Chef subnet name in machine bootstrap options

jkeiser:
This is based on #111 and chef/chef-provisioning#297.

### DIFF
--- a/lib/chef/provider/aws_provider.rb
+++ b/lib/chef/provider/aws_provider.rb
@@ -1,89 +1,22 @@
 require 'chef/provider/lwrp_base'
-require 'chef/provisioning/aws_driver/credentials'
 
 class Chef::Provider::AwsProvider < Chef::Provider::LWRPBase
   use_inline_resources
-
-  attr_reader :credentials
 
   # All these need to implement whyrun
   def whyrun_supported?
     true
   end
 
-  def from_hash hash
-    hash.each do |k,v|
-      begin
-        self.instance_variable_set("@#{k}", v)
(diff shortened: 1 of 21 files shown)


lib/chef/provider/aws_provider.rb
```diff
@@ -1,89 +1,34 @@
...
-  end
-
-  def sns
-    @sns ||= aws_init { AWS::SNS.new }
-  end
-
-  def sqs
-    @sqs ||= aws_init { AWS::SQS.new }
+  def new_driver
+    run_context.chef_provisioning.driver_for(new_resource.driver)
```
tyler-ball: Because we are accessing this frequently and it doesn't look like it will change, can we memoize this?

lib/chef/provider/aws_provider.rb
```diff
@@ -1,89 +1,34 @@
...
-  end
-
-  private
-  def aws_init
-    credentials = @credentials.default
-    region = new_resource.region_name || credentials[:region]
-    # Pivot region
-    AWS.config(:region => region)
-    yield
+  def get_subnet(subnet)
```
tyler-ball: It looks like you have this code duplicated in the driver inside `bootstrap_options_for` - can you move this method into the driver?

tyler-ball:
A few comments/questions, and this needs a rebase.  But the conversion to accessing the driver from the providers looks good!

jkeiser:
This is superceded entirely by #123 :)

PR #113, opened 2015-02-27
Add internet_gateway support and route to VPC

jkeiser:
Also add map_public_ip_on_launch support to subnet

This is based off #111.

### DIFF
--- a/lib/chef/provider/aws_subnet.rb
+++ b/lib/chef/provider/aws_subnet.rb
@@ -6,7 +6,10 @@ class Chef::Provider::AwsSubnet < Chef::Provider::AwsProvider
   action :create do
     fail "Can't create a Subnet without a CIDR block" if new_resource.cidr_block.nil?
 
-    if existing_subnet == nil
+    if existing_subnet
+      subnet = existing_subnet
+      # TODO update things
+    else
       converge_by "Creating new Subnet #{fqn} in VPC #{new_resource.vpc} in #{new_driver.aws_config.region}" do
         opts = { :vpc => vpc_id }
         opts[:availability_zone] = new_resource.availability_zone if new_resource.availability_zone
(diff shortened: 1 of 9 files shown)


lib/chef/provider/aws_subnet.rb
```diff
@@ -6,22 +6,40 @@ class Chef::Provider::AwsSubnet < Chef::Provider::AwsProvider
   action :create do
     fail "Can't create a Subnet without a CIDR block" if new_resource.cidr_block.nil?
 
-    if existing_subnet == nil
-      converge_by "Creating new Subnet #{fqn} in VPC #{new_resource.vpc} in #{new_resource.region_name}" do
+    if existing_subnet
+      subnet = existing_subnet
+      # TODO update things
```
tyler-ball: This seems necessary

lib/chef/provider/aws_subnet.rb
```diff
@@ -6,22 +6,40 @@ class Chef::Provider::AwsSubnet < Chef::Provider::AwsProvider
...
         new_resource.subnet_id subnet.id
         new_resource.save
       end
     end
+
+    subnet_desc = new_driver.ec2.client.describe_subnets(subnet_ids: [ subnet.id ])[:subnet_set].first
+    if new_resource.map_public_ip_on_launch
+      if !subnet_desc[:map_public_ip_on_launch]
+        converge_by "Turning on automatic public IPs for subnet #{subnet.id}" do
+          new_driver.ec2.client.modify_subnet_attribute(subnet_id: subnet.id, map_public_ip_on_launch: { value: true })
```
tyler-ball: Can't you set this attribute during the create?  http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/EC2/Client.html#create_subnet-instance_method says you can

lib/chef/provider/aws_vpc.rb
```diff
@@ -6,20 +6,70 @@ class Chef::Provider::AwsVpc < Chef::Provider::AwsProvider
   action :create do
     fail "Can't create a VPC without a CIDR block" if new_resource.cidr_block.nil?
 
-    if existing_vpc == nil
-      converge_by "Creating new VPC #{fqn} in #{new_resource.region_name}" do
+    if existing_vpc
+      vpc = existing_vpc
```
tyler-ball: This also seems like a pending TODO

tyler-ball: It seems like there is another TODO here to update the resource

lib/chef/provider/aws_subnet.rb
jkeiser: Can you actually set the map_public_ip_on_launch attribute on launch?  I didn't think that was possible: http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/EC2/SubnetCollection.html#create-instance_method

I'm not morally opposed to setting things on create, but I think the logic for update is identical to create in this case (the thing must already exist before you can set its attribute).

jkeiser:
Closing to reopen against master.

Issue #114, opened 2015-02-27
AWS::Core::OptionGrammar::FormatError updating a load balancer


christinedraper:
I'm running the recipe below (using 0.3 driver) to create a autoscaling group with a load balancer. It runs once, then fails on the second run, with the error below.

I suspect that on the second run, the `machines([])` statement is causing an attempt to deregister the instances in the autoscaling group, which isnt what I intend (but shouldnt fail).  I get a different error if I dont specify machines, so I'm not sure what the recipe should actually be (it did work with like this in 0.2.1).

```
require 'chef/provisioning/aws_driver'
with_driver('aws')

load_balancer "elbappserverelb" do
  load_balancer_options({:availability_zones=>["us-west-2c", "us-west-2b", "us-west-2a"], :listeners=>[{:instance_port=>3001, :protocol=>"HTTP", :instance_protocol=>"HTTP", :port=>3001}]})
  machines([])
end

aws_launch_config "asgappservergroup_config" do
  image "ami-9f3015af"
  instance_type "t1.micro"
end

aws_auto_scaling_group "asgappservergroup_group" do
  launch_config 'asgappservergroup_config'
  max_size(2)
  min_size(1)
  desired_capacity(1)
  load_balancers(["elbappserverelb"])
end

machine "instancedbserver" do
  add_machine_options({:bootstrap_options=>{:image_id=>"ami-f1ce8bc1", :instance_type=>"t1.micro", :disable_api_termination=>false, :kernel_id=>"aki-fc8f11cc", :key_name=>"test1_aws", :monitoring_enabled=>false}})
  run_list(["starter_for_ai-autoscaling"])
  tags([])
  attribute 'topo', {"node_type"=>"db server", "name"=>"Starter_for_ai-autoscaling"} 
end
```

```
AWS::Core::OptionGrammar::FormatError: expected string value for key instance_id of member 1 of option instances
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.61.0/lib/aws/core/option_grammar.rb:93:in `validate'
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.61.0/lib/aws/core/option_grammar.rb:346:in `block in validate'
<snip>
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/aws-sdk-v1-1.61.0/lib/aws/core/client.rb:476:in `client_request'
[...]

PR #117, opened 2015-02-28
Fix #116 and #114

### LINKED ISSUE #116: with_driver no longer getting region from profile
Using latest code in master (as of Saturday 28 Feb pm)

```
with_driver('aws')

load_balancer "elbappserverelb" do
  load_balancer_options({:availability_zones=>["us-west-2c", "us-west-2b", "us-west-2a"], :listeners=>[{:instance_port=>3001, :protocol=>"HTTP", :instance_protocol=>"HTTP", :port=>3001}]})
  machines []
end
```

Fails with  AWS::ELB::Errors::ValidationError: Invalid Availability Zone: us-west-2c

Even though us-west-2 is in the default profile.

with_driver 'aws::us-west-2' works.  with_driver 'aws:test1' doesnt.

christinedraper:
For #116, I added back the logic to set region to the profile in credentials if not specified on driver

For #114, I allowed a load balancer to be created when machines is not specified at all. In this case, it will not try to delete existing instances. This is needed so it can work with an autoscaling cluster resource that specifies a load balancer.

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -170,23 +170,25 @@ module AWSDriver
         end
       end
 
-      # Update instance list
+      # Update instance list, but only if there are machines specified
       actual_instance_ids = Set.new(actual_elb.instances.map { |i| i.instance_id })
 
-      instances_to_add = machine_specs.select { |s| !actual_instance_ids.include?(s.location['instance_id']) }
-      instance_ids_to_remove = actual_instance_ids - machine_specs.map { |s| s.location['instance_id'] }
-
-      if instances_to_add.size > 0
-        perform_action.call("  add machines #{instances_to_add.map { |s| s.name }.join(', ')}") do
-          instance_ids_to_add = instances_to_add.map { |s| s.location['instance_id'] }
-          Chef::Log.debug("Adding instances #{instance_ids_to_add.join(', ')} to load balancer #{actual_elb.name} in region #{aws_config.region}")
-          actual_elb.instances.add(instance_ids_to_add)
+      if machine_specs
+        instances_to_add = machine_specs.select { |s| !actual_instance_ids.include?(s.location['instance_id']) }
+        instance_ids_to_remove = actual_instance_ids - machine_specs.map { |s| s.location['instance_id'] }
+  
+        if instances_to_add.size > 0
+          perform_action.call("  add machines #{instances_to_add.map { |s| s.name }.join(', ')}") do
+            instance_ids_to_add = instances_to_add.map { |s| s.location['instance_id'] }
+            Chef::Log.debug("Adding instances #{instance_ids_to_add.join(', ')} to load balancer #{actual_elb.name} in region #{aws_config.region}")
+            actual_elb.instances.add(instance_ids_to_add)
+          end
         end
-      end
-
-      if instance_ids_to_remove.size > 0
-        perform_action.call("  remove instances #{instance_ids_to_remove}") do
-          actual_elb.instances.remove(instance_ids_to_remove)
+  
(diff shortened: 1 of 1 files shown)


lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -170,23 +170,25 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
...
 
-      instances_to_add = machine_specs.select { |s| !actual_instance_ids.include?(s.location['instance_id']) }
-      instance_ids_to_remove = actual_instance_ids - machine_specs.map { |s| s.location['instance_id'] }
-
-      if instances_to_add.size > 0
-        perform_action.call("  add machines #{instances_to_add.map { |s| s.name }.join(', ')}") do
-          instance_ids_to_add = instances_to_add.map { |s| s.location['instance_id'] }
-          Chef::Log.debug("Adding instances #{instance_ids_to_add.join(', ')} to load balancer #{actual_elb.name} in region #{aws_config.region}")
-          actual_elb.instances.add(instance_ids_to_add)
+      if machine_specs
```
jkeiser: Do you have the latest chef-provisioning?  I ask because I think I fixed this issue on the caller side: https://github.com/chef/chef-provisioning/blob/master/lib/chef/provider/load_balancer.rb#L28

christinedraper: I think what you've done caller side is turn "not specifying machines" into
"specifying NO machines" - i.e. equivalent of saying "machines([]) - right?

If so, the problem is that whenever the recipe runs, the load balancer will
try to detach the auto-scaling instances (triggering the AWS FormatError).
If we got past the error, then the aws-auto-scaling-group resource would
need to reattach the instances.

 So that's why I think we need a way to say "there needs to be a load
balancer, but dont do anything to its attached instances".

Regards,
Christine

christinedraper: Just verified the behavior with chef-provisioning 0.19 with no machines attribute... basically, on second run it removes the instances that were added to it by the auto-scaling group, and you get left with a load balancer with no instances.

jkeiser: ...

Damn straight.  I hadn't thought about that.  We should support patchy behavior whenever we can.  I will reverse that patch and we will be able to check this in.

jkeiser:
@christinedraper do me a favor and check out https://github.com/chef/chef-provisioning/pull/299 (the `jk/empty_lb` branch of chef-provisioning) and see if that brings things back into harmony?  Thanks!

jkeiser:
WAIT.  One sec.  That was way bigger than it should have been because I branched off the wrong branch.

jkeiser:
OK, that should do it.  Same branch, all is sane now :)

christinedraper:
Harmony restored - this PR + jk/empty_lb do the necessary

tyler-ball:
I don't think this is the logic we want - if you specify a load balancer as
```ruby
load_balancer "elbappserverelb" do
  load_balancer_options({:availability_zones=>["us-west-2c", "us-west-2b", "us-west-2a"], :listeners=>[{:instance_port=>3001, :protocol=>"HTTP", :instance_protocol=>"HTTP", :port=>3001}]})
end
```

Then we all agree that on a create it should not assign any machines to `elbappserverelb`.  But why then on an update should it not unassign any assigned instances?  It seems like create and update now have different behavior.

I do agree that we should be able to create a load_balancer with no machines tied to it - that is a separate bug that needs to be fixed (already got fixed?).

jkeiser:
@tyler-ball yep, that's the short of it :)

tyler-ball:
Validated this locally and successfully merged to master - thanks @christinedraper !

PR #118, opened 2015-03-01
Fix specs

erikvanbrakel:
Support for 'with_driver' and 'driver' (PR #111) broke the recently added specs. This seems to fix it, but is there a better way?

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -50,7 +50,8 @@ module AWSDriver
       @aws_config = AWS::Core::Configuration.new(
         access_key_id:     credentials[:aws_access_key_id],
         secret_access_key: credentials[:aws_secret_access_key],
-        region: region || credentials[:region]
+        region: region || credentials[:region],
+        stub_requests: AWS.config.stub_requests
       )
     end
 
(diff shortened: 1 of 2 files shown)


jkeiser:
@erikvanbrakel probably the best way is to modify the `run_context` at the top there, so you (hopefully) don't have to do that:

```ruby
run_context.chef_provisioning.with_driver 'aws'
```

I'm OK merging this too; let me know what you'd like to do :)

erikvanbrakel:
I'll change the run context. Was actually looking for that, but I am not that good at chef or ruby yet, so I couldn't find it ;-)

Op 2 mrt. 2015 19:09 schreef John Keiser <notifications@github.com>:

@erikvanbrakel<https://github.com/erikvanbrakel> probably the best way is to modify the run_context at the top there, so you (hopefully) don't have to do that:

run_context.chef_provisioning.with_driver 'aws'

I'm OK merging this too; let me know what you'd like to do :)

erikvanbrakel:
Decided to change the run_context where it's initialized, as it's also used by the provider. Probably cleanest option now.

erikvanbrakel:
Hmm, identified another issue with the specs, which is a blocker for this PR: AWS is not being stubbed anymore. I believe AWS.stub! sets a default config setting, but as we create a new config object in lib/chef/provisioning/driver.rb:50, this is ignored. Will try to fix this...

erikvanbrakel:
There, that should do it.

PR #127, opened 2015-03-11
Add DHCP options support

jkeiser:


tyler-ball:
Hey @jkeiser I've been adding some specs - will you take a look and see what you think?  I've got decent coverage on VPCs

spec/integration/aws_vpc_spec.rb
```diff
@@ -0,0 +1,331 @@
+require 'spec_helper'
```
jkeiser: I think the pattern looks good!  At some point we should allow it to talk to a real AWS as well as a fake one, though :)

PR #132, opened 2015-03-17
Make security groups idempotent, add better syntax

jkeiser:
```ruby
aws_security_group 'blah' do
  inbound_rules 'other-security-group' => 80 # allow other security groups to hit 80
  outbound_rules 80 => '0.0.0.0/0' # allow members of this security group to send traffic to port 80 of anything
end
```

### DIFF
--- a/docs/examples/ref.rb
+++ b/docs/examples/ref.rb
@@ -17,18 +17,14 @@ end
 
 aws_security_group 'ref-sg1' do
   vpc 'ref-vpc'
-  inbound_rules [ { ports: 22, protocol: :tcp, sources: [ '0.0.0.0/0' ] } ]
+  inbound_rules '0.0.0.0/0' => 22
 end
 
 aws_security_group 'ref-sg2' do
   vpc 'ref-vpc'
 
-  inbound_rules [
-    {:ports => 2223, :protocol => :tcp, :sources => ['ref-sg1'] }
-  ]
-  outbound_rules [
-    {:ports => 2223, :protocol => :tcp, :destinations => ['ref-sg1'] }
-  ]
+  inbound_rules 'ref-sg1' => 2224
+  outbound_rules 2224 => 'ref-sg1'
 end
 
 aws_route_table 'ref-public' do
--- a/docs/examples/sg_test.rb
+++ b/docs/examples/sg_test.rb
@@ -1,0 +1,100 @@
+with_driver 'aws'
+
+require 'chef/provisioning/aws_driver'
+
+vpc = aws_vpc 'test-vpc' do
+  cidr_block '10.0.0.0/24'
+  internet_gateway true
+end
+
+# Empty
+
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+  action :delete
+end
+
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+end
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+end
+
+# Things we can reference
+aws_security_group 'other-sg' do
+  vpc 'test-vpc'
+end
+
+aws_subnet 'test-subnet' do
+  vpc 'test-vpc'
+end
+
+recipe = self
+ruby_block 'lb' do
+  block do
+    recipe.load_balancer 'other-lb' do
+      load_balancer_options subnets: vpc.aws_object.subnets.map { |s| s }
+    end
+  end
+end
+
+# Add/update
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+  action :delete
+end
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+  inbound_rules '0.0.0.0/0'                   => 22,
+                'other-sg'                    => 1024..2048,
+                { load_balancer: 'other-lb' } => 1024..2048
+  outbound_rules 443        => '0.0.0.0/0',
+                 2048..4096 => 'other-sg',
+                 2048..4096 => { load_balancer: 'other-lb' }
+end
+
+# Add one inbound rule, change one inbound rule, add to one inbound rule
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+  inbound_rules '0.0.0.0/0' => 80,
+                'other-sg'  => [ 80, 1024..2048 ],
+                '127.0.0.1' => 1024..2048,
+                { load_balancer: 'other-lb' } => 1024..2048
+end
+
+# Add one outbound rule, change one outbound rule, add to one outbound rule
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+  outbound_rules 80                 => '0.0.0.0/0',
+                 [ 80, 2048..4096 ] => 'other-sg',
+                 2048..4096         => '127.0.0.1',
+                 1024..2048         => { load_balancer: 'other-lb' }
+end
+
+
+# Idempotence
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+  inbound_rules '0.0.0.0/0'                   => 80,
+                'other-sg'                    => [ 80, 1024..2048 ],
+                '127.0.0.1'                   => 1024..2048,
+                { load_balancer: 'other-lb' } => 1024..2048
+  outbound_rules 80                 => '0.0.0.0/0',
+                 [ 80, 2048..4096 ] => 'other-sg',
+                 2048..4096         => '127.0.0.1',
+                 1024..2048         => { load_balancer: 'other-lb' }
+end
+
+# Idempotence (alternate way of writing it)
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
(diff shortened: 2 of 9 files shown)


docs/examples/sg_test.rb
```diff
@@ -0,0 +1,100 @@
...
+
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+  action :delete
+end
+
+aws_security_group 'test-sg' do
+  vpc 'test-vpc'
+end
+aws_security_group 'test-sg' do
```
tyler-ball: Are you testing for idempotency with this duplicate `'test-sg'`?

jkeiser: Yep!

docs/examples/sg_test.rb
```diff
@@ -0,0 +1,100 @@
...
+aws_security_group 'other-sg' do
+  vpc 'test-vpc'
+end
+
+aws_subnet 'test-subnet' do
+  vpc 'test-vpc'
+end
+
+recipe = self
+ruby_block 'lb' do
```
tyler-ball: I assume you're putting this in a ruby-block to prevent the resource from being declared until converge time.  Why is this necessary?

lib/chef/provider/aws_security_group.rb
```diff
@@ -1,83 +1,288 @@
...
-
-        sg = driver.ec2.security_groups.create(new_resource.name, options)
-      end
+    if sg
+      update_security_group(sg)
+    else
+      sg = create_security_group
     end
 
     new_resource.save_managed_entry(sg, action_handler)
```
tyler-ball: In the [aws_vpc](https://github.com/chef/chef-provisioning-aws/blob/27f4b4b9c8532b80b87996e87861b77202bf4551/lib/chef/provider/aws_vpc.rb#L74) provider you're only saving the managed entry if we fall into the create block.  That prevents creating a managed entry if the user references the VPC by `vpc_id`.  Shouldn't we have the same behavior here?

lib/chef/provisioning/aws_driver/super_lwrp.rb
```diff
@@ -33,6 +33,12 @@ def self.attribute(attr_name, validation_opts={})
             super
           end
         end
+
+        # FUUUUUU cloning
```
tyler-ball: This seems like a dangerous change.  Is this intended to be permanent?

lib/chef/provider/aws_security_group.rb
```diff
@@ -1,83 +1,288 @@
...
+    # An array is always considered a list of actors.  Each one may follow any supported format.
+    when Array
+      actor_spec.map { |a| get_actors(vpc, a) }
+
+    # Hashes come in several forms:
+    when Hash
+      # The default AWS Ruby SDK form with :user_id, :group_id and :group_name forms
+      if actor_spec.keys.all? { |key| [ :user_id, :group_id, :group_name ].include?(key) }
+        if actor_spec.has_key?(:group_name)
+          actor_spec[:group_id] ||= vpc.security_groups.filter('group-name', actor_spec[:group_name]).first.id
```
tyler-ball: Because `vpc` isn't a required resource attribute, it could be `nil` here

lib/chef/resource/aws_security_group.rb
```diff
@@ -7,11 +7,45 @@ class Chef::Resource::AwsSecurityGroup < Chef::Provisioning::AWSDriver::AWSResou
   actions :create, :delete, :nothing
   default_action :create
 
-  attribute :name,          kind_of: String, name_attribute: true
-  attribute :vpc,           kind_of: [ String, AwsVpc, AWS::EC2::VPC ]
-  attribute :description,   kind_of: String
-  attribute :inbound_rules
-  attribute :outbound_rules
+  attribute :name,           kind_of: String, name_attribute: true
+  attribute :vpc,            kind_of: [ String, AwsVpc, AWS::EC2::VPC ]
```
tyler-ball: The provider assumes this will always be populated (and it is required in a create call to amazon, unless operating in ec2 classic mode) - can we make it `required: true`?

docs/examples/sg_test.rb
jkeiser: Because the VPC doesn't exist until converge time, so load_balancer_options can't set the subnet until then.

lib/chef/provisioning/aws_driver/super_lwrp.rb
jkeiser: Yep.  Not dangerous though.  All Cheffish and chef-provisioning objects have been doing this for a very long time, no issues yet.

The reason for it is, cloning kills patchiness.  We need to be able to declare multiple things with their own properties without polluting them with other resources' properties.

lib/chef/provider/aws_security_group.rb
```diff
@@ -1,83 +1,288 @@
...
+      if desired_vpc != sg.vpc_id
+        raise "Security Group VPC cannot be changed after being created!  Desired VPC for #{new_resource.name} (#{sg.id}) was #{new_resource.vpc} (#{desired_vpc}) and actual VPC is #{sg.vpc_id}"
+      end
+    end
+  end
+
+  def create_security_group
+    sg = nil
+    converge_by "create new SG #{new_resource.name} in #{region}" do
+      options = { description: new_resource.description }
```
tyler-ball: Does AWS get angry if this is nil?

docs/examples/sg_test.rb
tyler-ball: :+1:

docs/examples/sg_test.rb
tyler-ball: What about
```ruby
load_balancer 'other-lb' do
  load_balancer_options lazy {
    subnets: vpc.aws_object.subnets
  }
end
```

tyler-ball:
I would love to see some test coverage, but other than that this is looking great.  I'm going to test locally and see if I can find any lurking bugs!

docs/examples/sg_test.rb
```diff
@@ -0,0 +1,100 @@
+with_driver 'aws'
```
tyler-ball: This should be `with_driver 'aws::eu-west-1'`

lib/chef/provider/aws_security_group.rb
jkeiser: We absolutely should!

lib/chef/provider/aws_security_group.rb
jkeiser: Absolutely!  Will fix :)

lib/chef/resource/aws_security_group.rb
jkeiser: I'd actually rather not, because on *update* you don't have to care what VPC it's in.  We should just error out on create if you don't specify one.

tyler-ball: :+1:

lib/chef/provider/aws_security_group.rb
jkeiser: The new PR should address this.

docs/examples/ref.rb
```diff
@@ -17,18 +17,14 @@
...
   vpc 'ref-vpc'
 
-  inbound_rules [
-    {:ports => 2223, :protocol => :tcp, :sources => ['ref-sg1'] }
-  ]
-  outbound_rules [
-    {:ports => 2223, :protocol => :tcp, :destinations => ['ref-sg1'] }
-  ]
+  inbound_rules 'ref-sg1' => 2224
+  outbound_rules 2224 => 'ref-sg1'
```
dblessing: How is the protocol specified now? I'm assuming it defaults to tcp, which is fine, as long as it can be overridden in some way still?

lib/chef/provider/aws_security_group.rb
jkeiser: Nope.  It copies the name.  (See `ref.rb`)

docs/examples/sg_test.rb
jkeiser: I don't actually know if `load_balancer_options` accepts lazy.  We need to get lazy working through and through.

docs/examples/ref.rb
jkeiser: The old syntax still works; but here's how you do it in the new:

```ruby
outbound_rules { port_range: 2224, protocol: :udp } => 'ref-sg1'
inbound_rules 'ref-sg1' => { port_range: 2224, protocol: :udp }
```

Arrays work as well, on both sides:

```ruby
outbound_rules [22, 1024..2048] => [ '10.0.0.0/8', '169.0.0.0/24' ]
```

And you can specify the hashes in the ports section, as well.

lib/chef/provider/aws_security_group.rb
tyler-ball: I was lying here - vpc is passed in and it is provided from the AWS object, so it cannot be nil

PR #133, opened 2015-03-17
aws_ebs_volume add attach, detach, and ref tests

patrick-wright:
This is the latest version of adding the attach and detach actions based on the recent chef-provisioning-aws overhaul.

#### delete action
* if a volume is attached to any instance, it will be automatically detached before attempting the delete (otherwise aws will raise an exception)

#### attach action
* if a volume is :available, it will be attached
* if a volume is attached to another instance, it will be detached and attached to the new_resource
* if a volume is attached to the expected machine, but to a different device, it will be detached and attached to the new_resource
* if a volume doesn't exist at the time of attachment an exception is raised

#### detach action
* if a volume is :in_use, it will be detached
* if a volume doesn't exist at the time of attachment a warning is logged

#### wait_for_volume_status checks
* create, delete, attach, and detach call #wait_for_volume_status after converge_by blocks have executed.

Status waits:
* create => :available
* delete => !volume.exists?
* attach => :in_use
* detach => :available

#### Ref Tests
Example ref recipes have been added to test the resource.  These will eventually be added to the new provisioning integration test framework.
* _idempotence: verifies each action does not error when called twice to honor #wait_for_volume_status calls
* _reattachment: verifies all attach state combinations
* _pre_existing: verifies that an existing ebs volume can be referenced and then become managed my provisioning (automated to simulate existing volume)

@tyler-ball @jkeiser @chef/ociv

### DIFF
--- a/docs/examples/aws_ebs_volume_idempotence.rb
+++ b/docs/examples/aws_ebs_volume_idempotence.rb
@@ -25,13 +25,6 @@ ebs_volume.run_action(:detach) # up to date
 ebs_volume.run_action(:delete)
 ebs_volume.run_action(:delete) # up to date
 
-# TODO test elsewhere
-# aws_ebs_volume 'ebs-12345678' do
-#   action :attach
-#   machine 'ref-machine2'
-#   device '/dev/xvdf'
-# end
-
 machine ref_machine1.name do
   action :destroy
 end
@@ -39,4 +32,3 @@ end
 aws_key_pair 'ref-key-pair-ebs' do
   action :delete
 end
-
--- a/docs/examples/aws_ebs_volume_idempotence.rb
+++ b/docs/examples/aws_ebs_volume_idempotence.rb
@@ -2,28 +2,47 @@ require 'chef/provisioning/aws_driver'
 
 with_driver 'aws::us-west-2'
 
-key = aws_key_pair 'ref-key-pair-ebs'
-key.run_action(:create)
+aws_key_pair 'ref-key-pair-ebs'
 
 ref_machine1 = machine 'ref-machine-1'
-ref_machine1.run_action(:converge)
 
-ebs_volume = aws_ebs_volume 'ref-volume-ebs' do
-  action :nothing
-  machine ref_machine1.name
-  device '/dev/xvdf'
+aws_ebs_volume 'ref-volume-ebs' do
   availability_zone 'us-west-2a'
   size 1
 end
 
-ebs_volume.run_action(:create)
-ebs_volume.run_action(:create) # up to date
-ebs_volume.run_action(:attach)
-ebs_volume.run_action(:attach) # up to date
-ebs_volume.run_action(:detach)
-ebs_volume.run_action(:detach) # up to date
-ebs_volume.run_action(:delete)
-ebs_volume.run_action(:delete) # up to date
+aws_ebs_volume 'ref-volume-ebs' do
+  availability_zone 'us-west-2a'
+  size 1
+end
+
+aws_ebs_volume 'ref-volume-ebs' do
+  action :attach
+  machine ref_machine1.name
+  device '/dev/xvdf'
+end
+
+aws_ebs_volume 'ref-volume-ebs' do
+  action :attach
+  machine ref_machine1.name
+  device '/dev/xvdf'
+end
+
+aws_ebs_volume 'ref-volume-ebs' do
+  action :detach
+end
+
+aws_ebs_volume 'ref-volume-ebs' do
+  action :detach
+end
+
+aws_ebs_volume 'ref-volume-ebs' do
+  action :delete
+end
+
+aws_ebs_volume 'ref-volume-ebs' do
+  action :delete
+end
 
 machine ref_machine1.name do
   action :destroy
--- a/lib/chef/provider/aws_ebs_volume.rb
+++ b/lib/chef/provider/aws_ebs_volume.rb
@@ -1,8 +1,10 @@
 require 'chef/provisioning/aws_driver/aws_provider'
 require 'cheffish'
 require 'date'
+require 'chef/provisioning/aws_driver/mixin/aws_instance'
 
 class Chef::Provider::AwsEbsVolume < Chef::Provisioning::AWSDriver::AWSProvider
+  include Chef::Provisioning::Mixin::AWSInstance
 
   action :create do
     aws_object = new_resource.aws_object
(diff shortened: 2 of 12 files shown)


docs/examples/aws_ebs_volume_idempotence.rb
```diff
@@ -0,0 +1,34 @@
...
+end
+
+ebs_volume.run_action(:create)
+ebs_volume.run_action(:create) # up to date
+ebs_volume.run_action(:attach)
+ebs_volume.run_action(:attach) # up to date
+ebs_volume.run_action(:detach)
+ebs_volume.run_action(:detach) # up to date
+ebs_volume.run_action(:delete)
+ebs_volume.run_action(:delete) # up to date
```
tyler-ball: Just talked with @jkeiser - we're going to disable resource cloning for all the aws_* resources (https://github.com/chef/chef-provisioning-aws/blob/jk/awesome_security_groups/lib/chef/provisioning/aws_driver/super_lwrp.rb#L37-L40).  So you should be able to update this to look like:

```ruby
aws_key_pair 'ref-key-pair-ebs'

machine 'ref-machine-1' do
  machine_options bootstrap_options: { key_name: 'ref-key-pair-ebs' }
end

aws_ebs_volume 'ref-volume-ebs' do
  action :nothing
  machine ref_machine1.name
  device '/dev/xvdf'
  availability_zone 'us-west-2a'
  size 1
end

aws_ebs_volume 'ref-volume-ebs' do
  action :create
end
# up to date
aws_ebs_volume 'ref-volume-ebs' do
  action :create
end
aws_ebs_volume 'ref-volume-ebs' do
  action :attach
end
# up to date
aws_ebs_volume 'ref-volume-ebs' do
  action :attach
end
aws_ebs_volume 'ref-volume-ebs' do
  action :detach
end
# up to date
aws_ebs_volume 'ref-volume-ebs' do
  action :detach
end
aws_ebs_volume 'ref-volume-ebs' do
  action :delete
end
# up to date
aws_ebs_volume 'ref-volume-ebs' do
  action :delete
end
```

This will also prevent these actions from running at compile time.

tyler-ball: Although now that I think about it, you should be able to do my previous code block now and have it work without resource cloning.  What was the problem you were seeing when trying to do that?

patrick-wright: Before all the recent provisioning changes, only the first compiled resource would be executed.  However, I just tried it from master and it worked!  I'll b make the changes to the other tests as well.

lib/chef/resource/aws_ebs_volume.rb
```diff
@@ -20,6 +20,10 @@ class Chef::Resource::AwsEbsVolume < Chef::Provisioning::AWSDriver::AWSResourceW
     name =~ /^vol-[a-f0-9]{8}$/ ? name : nil
   }
 
+  attribute :machine,           kind_of: String
```
tyler-ball: This can be `attribute :machine, kind_of [String, AwsMachine, AWS::EC2::Instance]`

To fetch the correct machine from any of the 3 provided types you use `Chef::Resource::AwsInstance.get_aws_object(new_resource.machine, resource: new_resource)` from inside your provider.

Some examples of this in other resources:
* https://github.com/chef/chef-provisioning-aws/blob/cad5c92b4a8fb3b616e45ccccc7fb7bb45af5b2d/lib/chef/provider/aws_security_group.rb#L248-L266
* https://github.com/chef/chef-provisioning-aws/blob/cad5c92b4a8fb3b616e45ccccc7fb7bb45af5b2d/lib/chef/provisioning/aws_driver/driver.rb#L603

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -37,12 +42,115 @@ class Chef::Provider::AwsEbsVolume < Chef::Provisioning::AWSDriver::AWSProvider
...
+  action :detach do
+    aws_object = new_resource.aws_object
+    status = aws_object ? aws_object.status : nil
+    case status
+    when nil
+      Chef::Log.warn "EBS volume #{new_resource.name} does not currently exist!"
+    when :in_use
+      detach
+    end
+    new_resource.save_managed_entry(aws_object, action_handler)
```
tyler-ball: I believe you only need this `save_managed_entry` call on the `action :create`.  The managed entry only contains AWS identifiable information so we don't need to save it on all actions.  We also don't want to create a managed entry if the user has supplied `volume_id` in the resource - because we have the unique identifier in that situation we don't need the managed entry.

tyler-ball:
This is looking really good!  I added 1 comment and I would like to see some tests added, but otherwise I think its almost done.

For an example of the tests I am writing, check out https://github.com/chef/chef-provisioning-aws/blob/master/spec/integration/aws_vpc_spec.rb - I'm also working on some improvments to that to remove some unecessary boilerplate.

tyler-ball:
The travis failure is because the [Gemfile](https://github.com/chef/chef-provisioning-aws/blob/93127e73bf96b1d0011ea7aa390c17f3ac4cac4c/Gemfile#L5) you checked in references a local path.  Update this to reference the git master branch and that should go away.

patrick-wright:
@tyler-ball here are my changes:
* Removed #save_managed_entry lines from :attach and :detach
* Added multi-machine types to machine attribute
 * Added a new recipe to validate they work.  I had issues with AwsInstance, see my comment
* Fixed Gemfile
* Have not started on specs

tyler-ball:
Yeah, I have also been having issues with the `get_aws_object` myself.  I have failing tests in https://github.com/chef/chef-provisioning-aws/pull/132 that highlight this.

I don't think you need the lazy blocks (when adding the machine) - at least, you shouldn't.  If you do, that is a bug on our side.

tyler-ball:
Actually @jkeiser - should we need the lazy blocks https://github.com/chef/chef-provisioning-aws/blob/pw/ebs_vol_attachment/docs/examples/aws_ebs_volume_machine_types.rb#L36 and https://github.com/chef/chef-provisioning-aws/blob/pw/ebs_vol_attachment/docs/examples/aws_ebs_volume_machine_types.rb#L44-L47 ?  Maybe I am wrong and they are necessary.

```ruby
m = machine 'test'
ebs_volume 'test_volume' do
  machine m
end
```
I thought ^^^ would work but maybe the reason my failing security groups test is because the machine doesn't have an aws object until convergence time, and the ebs_volume resource is trying to look it up at compile time.

patrick-wright:
m  = machine 'blah' returns the Chef::Resource::Machine

I bet
i = aws_instance 'blah' is what we want

PR #136, opened 2015-03-19
Standardize create/update/delete

jkeiser:
- Ensure data bag is associated with ID even if not creating new thing
- Handle other cases where ID is not the same as data bag


docs/examples/auto_scaling_cleanup.rb
```diff
@@ -2,10 +2,10 @@
 
 with_driver 'aws::eu-west-1' do
   aws_auto_scaling_group 'my-awesome-auto-scaling-group' do
-    action :delete
+    action :destroy
```
tyler-ball: :+1: for consistency (although this will break all existing cookbooks... again).  Can we just alias `:delete` to `:destroy` in the SuperLWRP so people have an easier migration?

lib/chef/provisioning/aws_driver/aws_resource.rb
```diff
@@ -5,6 +5,9 @@
 # Common AWS resource - contains metadata that all AWS resources will need
 module Chef::Provisioning::AWSDriver
 class AWSResource < Chef::Provisioning::AWSDriver::SuperLWRP
+  actions :create, :destroy, :nothing
```
tyler-ball: How will this work with `:attach` and `:detach` for @patrick-wright's ebs_volume PR?

patrick-wright: John and I talked about it in length. I'd be happy to give you a run down
tommorow.

tyler-ball: I just talked to him about it now, and I think I'm up to speed.  Nouns, not verbs!

lib/chef/provisioning/aws_driver/aws_provider.rb
```diff
@@ -22,8 +23,184 @@ def region
...
+
+          #
+          # We have both a desired ID and an entry ID.  Find out whether they
+          # match and warn if they don't (because we're going to reassociate and
+          # update the *desired* AWS thing.).
+          #
+          if desired_driver.driver_url == entry_driver.driver_url && desired_id == entry_id
+            Chef::Log.debug "#{new_resource.to_s} is already associated with #{entry_id} in #{entry_driver.driver_url}"
+          else
+            Chef::Log.warn "#{new_resource.to_s} is currently associated with #{entry_id} in #{entry_driver.driver_url}, but the desired ID is #{desired_id} in #{new_resource.driver.driver_url}!  Will associate with new desired ID #{desired_id}."
```
tyler-ball: There are a bunch of `Chef::Log.warn` entries around here that say the ID is going to change.  Where is the implementation of that?  Does that happen in all the `update_aws_object` methods defined in the individual providers?

lib/chef/provisioning/aws_driver/aws_provider.rb
```diff
@@ -22,8 +23,184 @@ def region
...
+    desired_driver = new_resource.driver
+    desired_id = new_resource.public_send(new_resource.class.aws_id_attribute) if new_resource.class.aws_id_attribute
+
+    #
+    # If the user specified an ID, delete THAT; do NOT delete the associated object.
+    #
+    if desired_id
+      aws_object = new_resource.class.get_aws_object(desired_id, resource: new_resource, required: false)
+      if aws_object
+        Chef::Log.debug "#{new_resource.to_s} provided #{new_resource.class.aws_id_attribute} #{desired_id} in #{desired_driver.driver_url}.  Will delete."
```
tyler-ball: I think you want `(#{desired_id})` to be consistent with your old log messages (id in parens)

lib/chef/provisioning/aws_driver/aws_provider.rb
```diff
@@ -22,8 +23,184 @@ def region
...
+  def create_aws_object
+    raise NotImplementedError, :create_aws_object
+  end
+
+  def update_aws_object(obj)
+    raise NotImplementedError, :update_aws_object
+  end
+
+  def destroy_aws_object(obj)
+    raise NotImplementedError, :destroy_aws_object
```
tyler-ball: It seems like this could actually have a default implementation of 
```ruby
converge_by "Destroying #{obj.class} #{obj.name} (#{obj.aws_id})" do
  obj.destroy
end
```

lib/chef/provisioning/aws_driver/aws_provider.rb
jkeiser: The ID changes in save_managed_entry, a bit below.  `desired_id` and `desired_driver` will be used to update `entry`.

lib/chef/provisioning/aws_driver/aws_provider.rb
jkeiser: Indeed!  Nice catch.

lib/chef/provisioning/aws_driver/aws_resource_with_entry.rb
```diff
@@ -2,21 +2,78 @@
...
   end
 
   protected
 
+  #
+  # Update an existing ManagedEntry object.
+  #
+  # @return true if the entry was changed, and false if not
+  #
+  def update_managed_entry(aws_object, managed_entry)
```
tyler-ball: I think we can achieve FULL unit test coverage of this method.  If we add new fields to the managed entry, we can add unit tests that show an old managed entry is correctly updated with the new fields.

lib/chef/provider/aws_vpc.rb
```diff
@@ -37,45 +32,20 @@ class Chef::Provider::AwsVpc < Chef::Provisioning::AWSDriver::AWSProvider
...
-  private
-
-  def create_vpc
-    vpc = nil
     converge_by "create new VPC #{new_resource.name} in #{region}" do
-      opts = { }
-      opts[:instance_tenancy] = new_resource.instance_tenancy if new_resource.instance_tenancy
-      vpc = driver.ec2.vpcs.create(new_resource.cidr_block, opts)
+      vpc = new_resource.driver.ec2.vpcs.create(new_resource.cidr_block, options)
       vpc.tags['Name'] = new_resource.name
```
tyler-ball: Future note: All the tag-able AWS objects include the module [AWS::EC2::TaggedItem](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/EC2/TaggedItem.html)

PR #142, opened 2015-03-21
aws_ebs_volume (jk/based on create_update_delete branch)

patrick-wright:
refactor to support "noun" resource patterns.

### DIFF
--- a/docs/examples/aws_ebs_volume.rb
+++ b/docs/examples/aws_ebs_volume.rb
@@ -25,14 +25,10 @@ aws_ebs_volume 'ref-volume-ebs' do
   device '/dev/xvdg'
 end
 
-# TODO: Fix validation issue
-# Chef::Exceptions::ValidationFailed
-#    ----------------------------------
-#    Option name must be a kind of String!  You passed false.
-# # detach
-# aws_ebs_volume 'ref-volume-ebs' do
-#   machine false
-# end
+# detach
+aws_ebs_volume 'ref-volume-ebs' do
+  machine false
+end
 
 # delete volume
 aws_ebs_volume 'ref-volume-ebs' do
--- a/docs/examples/aws_ebs_volume.rb
+++ b/docs/examples/aws_ebs_volume.rb
@@ -6,21 +6,43 @@ aws_key_pair 'ref-key-pair-ebs'
 
 with_machine_options :bootstrap_options => { :key_name => 'ref-key-pair-ebs' }
 
-#machine 'ref-machine-1'
+machine 'ref-machine-1'
 
+# create volume
 aws_ebs_volume 'ref-volume-ebs' do
   availability_zone 'us-west-2a'
   size 1
 end
 
+# attach to machine
+aws_ebs_volume 'ref-volume-ebs' do
+  machine 'ref-machine-1'
+  device '/dev/xvdf'
+end
+
+# reattach to different device
+aws_ebs_volume 'ref-volume-ebs' do
+  device '/dev/xvdg'
+end
+
+# TODO: Fix validation issue
+# Chef::Exceptions::ValidationFailed
+#    ----------------------------------
+#    Option name must be a kind of String!  You passed false.
+# detach
+aws_ebs_volume 'ref-volume-ebs' do
+  machine false
+end
+
+# delete volume
 aws_ebs_volume 'ref-volume-ebs' do
   action :destroy
 end
 
-# machine 'ref-machine-1' do
-#   action :destroy
-# end
+machine 'ref-machine-1' do
+  action :destroy
+end
 
-# aws_key_pair 'ref-key-pair-ebs' do
-#   action :destroy
-# end
+aws_key_pair 'ref-key-pair-ebs' do
+  action :destroy
+end
--- a/docs/examples/aws_ebs_volume.rb
+++ b/docs/examples/aws_ebs_volume.rb
@@ -7,6 +7,11 @@ aws_key_pair 'ref-key-pair-ebs'
 with_machine_options :bootstrap_options => { :key_name => 'ref-key-pair-ebs' }
 
 machine 'ref-machine-1'
+machine 'ref-machine-2'
+
+# machine_batch do
+#   machines 'ref-machine-1', 'ref-machine-2'
+# end
 
 # create volume
 aws_ebs_volume 'ref-volume-ebs' do
@@ -25,17 +30,40 @@ aws_ebs_volume 'ref-volume-ebs' do
   device '/dev/xvdg'
 end
 
+# reattach to different machine
+aws_ebs_volume 'ref-volume-ebs' do
+  machine 'ref-machine-2'
+  device '/dev/xvdf'
+end
+
+# skip reattachment attempt
+aws_ebs_volume 'ref-volume-ebs' do
+  machine 'ref-machine-2'
+  device '/dev/xvdf'
+end
+
+# create and attach
+aws_ebs_volume 'ref-volume-ebs-2' do
+  availability_zone 'us-west-2a'
+  size 1
+  machine 'ref-machine-1'
+  device '/dev/xvdf'
+end
+
 # detach
 aws_ebs_volume 'ref-volume-ebs' do
   machine false
 end
 
-# delete volume
-aws_ebs_volume 'ref-volume-ebs' do
-  action :destroy
-end
+# delete volumes
+['ref-volume-ebs', 'ref-volume-ebs-2'].each { |volume|
+  aws_ebs_volume volume do
+    action :destroy
+  end
+}
 
-machine 'ref-machine-1' do
+machine_batch do
+  machines 'ref-machine-1', 'ref-machine-2'
   action :destroy
 end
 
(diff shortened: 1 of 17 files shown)


lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -68,4 +99,97 @@ def desired_options
...
+  def update_attachment(volume)
+    status = volume.status
+    #
+    # If we were told to attach the volume to a machine, do so
+    #
+    if desired_instance.is_a?(AWS::EC2::Instance)
+      case status
+      when :in_use
+        attachment = volume.attachments.first
+        # wrong instance
```
patrick-wright: these conditions don't need to be this accurate.  just check whether instance and device both match; else just use the `detach(volume, :instance => attachment.instance, :device => attachment.device)` to detach.

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -68,4 +99,97 @@ def desired_options
...
+        # wrong device only
+        elsif attachment.instance == desired_instance && attachment.device != new_resource.device
+          detach(volume, :device => attachment.device)
+          attach(volume)
+        end
+
+      when :available
+        attach(volume)
+      when nil
+        raise "EBS volume #{new_resource.name} does not currently exist!"
```
patrick-wright: Should create a common message format
"EBS volume <name> (<volume_id>)" or something similiar

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -47,13 +61,30 @@ def update_aws_object(volume)
...
+      detach(volume, :instance => attachment.instance, :device => attachment.device)
+      delete(volume)
+    else
+      delete(volume)
     end
   end
 
   private
 
+  def desired_instance
```
patrick-wright: rename to attachment_instance?  "desired" is odd

lib/chef/resource/aws_ebs_volume.rb
```diff
@@ -5,6 +5,8 @@ class Chef::Resource::AwsEbsVolume < Chef::Provisioning::AWSDriver::AWSResourceW
 
   attribute :name,    kind_of: String, name_attribute: true
 
+  attribute :machine#,           kind_of: [ String, FalseClass ] TODO Fix why false still being validated as String (even without kind_of:)
```
patrick-wright: do we still need to add AwsInstance or AWS::EC2::Instance as machine types?

lib/chef/provider/aws_ebs_volume.rb
tyler-ball: If you find you're raising this multiple times, its probably better to just create a custom error:

```ruby
class EbsVolumeNotFound < RuntimeError do
  def initialize(new_resource) do
    super("EBS volume #{new_resource.name} (#{new_resource.volume_id}) does not exist")
  end
end
```

lib/chef/resource/aws_ebs_volume.rb
tyler-ball: Yes

lib/chef/resource/aws_ebs_volume.rb
```diff
@@ -17,9 +19,11 @@ class Chef::Resource::AwsEbsVolume < Chef::Provisioning::AWSDriver::AWSResourceW
     name =~ /^vol-[a-f0-9]{8}$/ ? name : nil
   }
 
+  attribute :device,            kind_of: String
+
   def aws_object
     driver, id = get_driver_and_id
     result = driver.ec2.volumes[id] if id
-    result && result.exists? && ![:destroyd, :deleting].include?(result.status) ? result : nil
+    result && result.exists? && ![:deleted, :deleting].include?(result.status) ? result : nil
```
tyler-ball: If an ebs volume has a status of `:deleted` or `:deleting`, will it eventually remove itself completely from AWS?  Thats how instances seem to work.  If volumes work the same way this seems good.

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -47,13 +61,30 @@ def update_aws_object(volume)
   end
 
   def destroy_aws_object(volume)
-    converge_by "delete EBS volume #{new_resource.name} in #{region}" do
-      volume.delete
+    status = volume.status
+    case status
```
tyler-ball: This can be DRYed to
```ruby
if status == :in_use
  attachment = volume.attachments.first
  detach(volume, :instance => attachment.instance, :device => attachment.device)
end
delete(volume)
```

lib/chef/provider/aws_ebs_volume.rb
tyler-ball: We should also overwrite either `to_s` or `inspect` on `AwsResource` to be `"#{new_resource.name} (#{new_resource.send(new_resource.aws_id_attribute)})"`

docs/examples/aws_ebs_volume.rb
```diff
@@ -0,0 +1,48 @@
...
+  machine 'ref-machine-1'
+  device '/dev/xvdf'
+end
+
+# reattach to different device
+aws_ebs_volume 'ref-volume-ebs' do
+  device '/dev/xvdg'
+end
+
+# TODO: Fix validation issue
```
tyler-ball: I was dealing with this same issue earlier - I think there is an error in the LWRPBase.  Can you post your full stacktrace for that `Option name must be a kind of String!` error?

lib/chef/provider/aws_ebs_volume.rb
tyler-ball: Maybe `desired_attachment`?  The `desired` simply signifies current vs desired state.  The ivar stores what machine we expect to be attached to the volume once convergence is finished.  Does that make sense?

lib/chef/provider/aws_ebs_volume.rb
tyler-ball: This comment seems valid - whether the volume is attached to the wrong instance, or has the wrong device name (/dev/sda1 vs /dev/sda2) we need to update the attachment.  This means calling detach and then attach, correct?

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -68,4 +99,97 @@ def desired_options
...
+
+    #
+    # If we were told to set the machine to false, detach it.
+    #
+    else
+      case status
+      when nil
+        Chef::Log.warn "EBS volume #{new_resource.name} does not currently exist!"
+      when :in_use
+        detach(volume, :instance => Chef::Resource::AwsInstance.get_aws_object(new_resource.machine, resource: new_resource))
```
tyler-ball: I think this is where your `option name cannot be false` error is coming from.  `new_resource.machine` is `false` in this case, so you're trying to lookup the aws object named `false` which doesn't work.  You will need to pull the machine name from the current attachments.

lib/chef/provider/aws_ebs_volume.rb
jkeiser: If you do, it needs to be more like aws_vpc[myvpc] (vpc-1234234) -- right now, resource.to_s does the former format.

tyler-ball: I added it in https://github.com/chef/chef-provisioning-aws/commit/0ab222127195b536d936a551a001b7eef257450d - my output locally looks like

```
  * aws_vpc[test-vpc] (no AWS object id) action create
    - create new VPC test-vpc in us-west-2
    - create data bag item test-vpc at http://127.0.0.1:8889
    -   add reference = {"id"=>"vpc-16f34e73"}
    -   add driver_url = "aws::us-west-2"
  * aws_security_group[test-sg] (no AWS object id) action create
    - Creating new SG test-sg in us-west-2
    - create data bag item test-sg at http://127.0.0.1:8889
    -   add reference = {"id"=>"sg-5e78453b"}
    -   add driver_url = "aws::us-west-2"
  * aws_security_group[test-sg] (sg-5e78453b) action destroy
    - Deleting SG test-sg in us-west-2
    - delete data bag item test-sg at http://127.0.0.1:8889
  * aws_vpc[test-vpc] (vpc-16f34e73) action destroy
    - delete VPC test-vpc (vpc-16f34e73) in us-west-2
    - delete data bag item test-vpc at http://127.0.0.1:8889
```

lib/chef/provider/aws_ebs_volume.rb
patrick-wright: I understand the intent.  I was referring to the definition of the word: "a strong feeling of wanting to have something or wishing for something to happen".  The word implies an emotional context, and made me laugh of bit.  I was looking for a more objective word to describe the intent :)

lib/chef/provider/aws_ebs_volume.rb
patrick-wright: perfect!

patrick-wright: I ran into an issue if `get_driver_and_id` is not defined like on aws_key_pair.  

I did this. Better way?
```ruby
  def to_s
    id = nil
    id = get_driver_and_id[1] if self.respond_to?(:get_driver_and_id)
    "#{declared_type}[#{@name}] (#{ id ? id : 'no AWS object id'})"
  end
```

patrick-wright:
@jkeiser @tyler-ball PR is ready for another review

PR #151, opened 2015-03-25
Feature: Propagate Virtual Private Gateway Routes for `aws_route_table` resource

dblessing:
Propagate routes from one or more virtual private gateways.

```
aws_route_table 'private_route_table' do
  ...
  vgw_route_propagation ['vgw-abcd1234']
end
```

### DIFF
--- a/lib/chef/provider/aws_route_table.rb
+++ b/lib/chef/provider/aws_route_table.rb
@@ -9,6 +9,8 @@ class Chef::Provider::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSProvider
     if !new_resource.routes.nil?
       update_routes(vpc, route_table)
     end
+
+    update_virtual_private_gateways(route_table, new_resource.virtual_private_gateways)
   end
 
   protected
@@ -85,6 +87,30 @@ class Chef::Provider::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSProvider
     current_routes.values.each do |current_route|
       action_handler.perform_action "remove route sending #{current_route.destination_cidr_block} to #{current_route.target.id}" do
         current_route.delete
+      end
+    end
+  end
+
+  def update_virtual_private_gateways(route_table, gateway_ids)
+    current_propagating_vgw_set = route_table.client.describe_route_tables(route_table_ids: [route_table.id]).route_table_set.first.propagating_vgw_set
+
+    # Add propagated routes
+    if gateway_ids
+      gateway_ids.each do |gateway_id|
+        if !current_propagating_vgw_set.reject! { |vgw_set| vgw_set[:gateway_id] == gateway_id }
+          action_handler.perform_action "enable route propagation for route table #{route_table.id} to virtual private gateway #{gateway_id}" do
+            route_table.client.enable_vgw_route_propagation(route_table_id: route_table.id, gateway_id: gateway_id)
+          end
+        end
+      end
+    end
+
+    # Delete anything that's left
+    if current_propagating_vgw_set
+      current_propagating_vgw_set.each do |vgw_set|
+        action_handler.perform_action "disabling route propagation for route table #{route_table.id} from virtual private gateway #{vgw_set[:gateway_id]}" do
+          route_table.client.disable_vgw_route_propagation(route_table_id: route_table.id, gateway_id: vgw_set[:gateway_id])
+        end
       end
     end
   end
--- a/lib/chef/resource/aws_route_table.rb
+++ b/lib/chef/resource/aws_route_table.rb
@@ -35,6 +35,17 @@ class Chef::Resource::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSResource
   attribute :vpc,    kind_of: [ String, AwsVpc, AWS::EC2::VPC ], required: true
 
   #
+  # Enable route propagation from one or more virtual private gateways
+  #
+  # The value should be an array of virtual private gateway ID:
+  # ```ruby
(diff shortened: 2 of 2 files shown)


dblessing:
@jkeiser Sorry to ping directly. Is there anything on my end holding up my various PRs? Since I've never had anything accepted into this project previously I wasn't sure about CLA's or whatever. I believe I've signed the Chef CLA, though.

jkeiser:
Nope!  I'll work on unblocking the pileup shortly.

tyler-ball:
For reference, I don't know anything about Virtual Private Gateways.  But I'm in the AWS console playing around with them and it looks like the action you can take is to attach or detach them from the VPC.  Does that have anything to do with propagating the routes to the route table?

lib/chef/resource/aws_route_table.rb
```diff
@@ -35,6 +35,16 @@ class Chef::Resource::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSResource
...
 
   #
+  # Enable route propagation from one or more virtual private gateways
+  #
+  # The value should be an array of virtual private gateway ID, like so:
+  # ```ruby
+  # vgw_route_propagation ['vgw-abcd1234', 'vgw-abcd5678']
+  # ```
+  #
+  attribute :vgw_route_propagation, kind_of: Array
```
tyler-ball: I could see us creating resources for `aws_virtual_private_gateway` and `aws_customer_gateway` in the future.  Because of this I would like this attribute renamed to `:virtual_private_gateway`.  Then the attribute would be named the same for route tables and VPCs.

For niceness, you should also have `kind_of: [String, Array]` - If they provide a string you can just turn it into an array with `virtual_private_gateway = [new_resource.virtual_private_gateway].flatten` in the provider.

tyler-ball:
The idempotent logic looks good, but you should update the [aws_route_table_spec](https://github.com/chef/chef-provisioning-aws/blob/master/spec/integration/aws_route_table_spec.rb) with tests for this.  Because these integration tests don't currently get ran through travis, you'll have to run them locally.

dblessing:
@tyler-ball 

Yes, the gateway must be attached. Then in the console go to Route Tables -> Select a route table -> Route Propagation tab and click edit. Check the box for any gateway that should propagate routes.

I'll take a look at your other suggestions. Thanks for the feedback.

dblessing:
I came back to look at tests. Integration tests actually require an AWS account that it can run against, right? I don't think I'll be able to run that locally because I don't have an account that I can let that stuff run in. That's kind of a large hurdle...

dblessing:
@tyler-ball Can you explain how I might test this? Since gateways currently can't be created via Chef I am not sure how to create one, with the static routes, and then assign to a route table in specs.

tyler-ball:
@dblessing Hmmm, you're right - its going to be hard to test this without being able to manage virtual private gateways.  Can you add an issue to chef-provisioning-aws to create a `aws_virtual_private_gateway` resource and in the issue reference this comment so we can remember we need to add tests for this?  If you do that, I'm :+1: on this issue and will merge it.

lib/chef/provider/aws_route_table.rb
```diff
@@ -89,6 +91,30 @@ def update_routes(vpc, route_table)
     end
   end
 
+  def update_virtual_private_gateways(route_table, gateway_ids)
+    current_propagating_vgw_set = route_table.client.describe_route_tables(route_table_ids: [route_table.id]).route_table_set.first.propagating_vgw_set
```
tyler-ball: we can't just call `route_table.propagating_vgw_set` here?

lib/chef/resource/aws_route_table.rb
```diff
@@ -35,6 +35,16 @@ class Chef::Resource::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSResource
...
 
   #
+  # Enable route propagation from one or more virtual private gateways
+  #
+  # The value should be an array of virtual private gateway ID:
+  # ```ruby
+  # virtual_private_gateways ['vgw-abcd1234', 'vgw-abcd5678']
+  # ```
+  #
+  attribute :virtual_private_gateways, kind_of: [ String, Array ]
```
tyler-ball: To make sure this is an array when in the provider you can add
```ruby
attribute :virtual_private_gateways, kind_of: [ String, Array ], coerce: proc { |v| [v].flatten }
```

`coerce` is defined at https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/super_lwrp.rb#L15 and is only supported for the AWS resources (currently).  But its super handy!

lib/chef/provider/aws_route_table.rb
dblessing: Not according to their API docs, but they could be lacking. I'll try it and see.

lib/chef/resource/aws_route_table.rb
dblessing: Very handy! Thanks.

lib/chef/provider/aws_route_table.rb
dblessing: Yep, `undefined method` when I try that.

dblessing:
@tyler-ball Issue created. This is good to go. Thanks for your feedback.

lib/chef/provider/aws_route_table.rb
tyler-ball: :disappointed: <- this is my sad face.  Thanks for trying!

tyler-ball:
@dblessing Thank you for this improvement!

PR #152, opened 2015-03-25
Better AWS tests

jkeiser:
This provides:
- Much better failure text when a Chef run fails
- Automatic AWS object cleanup when the test run exits
- `action :purge` support for VPC (delete all things in the VPC)
- Deep match support for objects, with better printing (DeepMatch).


lib/chef/provider/aws_vpc.rb
```diff
@@ -55,6 +55,17 @@ def update_aws_object(vpc)
   end
 
   def destroy_aws_object(vpc)
+    if purging
```
patrick-wright: @jkeiser when you attempt to delete a vpc from the UI the user is prompted:
```
Are you sure you want to delete this VPC?. Deleting this VPC will also delete objects associated with this VPC in this region.

Subnets
Security Groups
Network ACLs
VPN Attachments
Internet Gateways
Route Tables
Network Interfaces
VPC Peering Connections
	
Delete VPN Connection when deleting the VPC? (checkbox)
```

We can safely add additional vpc objects to the purge.  If there are running instances networked to the vpc being deleted will the api call not execute? Purge is an excellent work choice if instances must be terminated as well.

jkeiser:
CC @tyler-ball @fnichol

lib/chef/provider/aws_vpc.rb
```diff
@@ -55,6 +55,41 @@ def update_aws_object(vpc)
...
+      # TODO we need to destroy the associated data bags as well.
```
tyler-ball: Shouldn't we wait to delete the data bag until after the instance has successfully been deleted (next code block below)?  Also, this needs to be done.

Issue #145, opened 2015-03-24
Possibly lazy load credentials


dblessing:
Would it be possible to change credentials to be lazy loaded instead of looking for `~/.aws/config|credentials` on `aws_driver` load? The issue I'm running in to is the recipe is attempting to create `~/.aws/config` but `require 'chef/provisioning/aws_driver'` happens at compile time before the config file created. Thus, a failure occurs.

tyler-ball:
@dblessing thanks for filing this.  Do you have a workaround right now?  If not, you should be able to force your template/file resource to converge during compile time.

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #156, opened 2015-03-26
Allow AwsDriver to work even if you haven't explicitly required 'chef/pr...

jkeiser:
...ovisioning/aws_driver'

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -10,8 +10,11 @@ require 'chef/provisioning/machine/windows_machine'
 require 'chef/provisioning/machine/unix_machine'
 require 'chef/provisioning/machine_spec'
 
-require 'chef/provider/aws_key_pair'
 require 'chef/resource/aws_key_pair'
+require 'chef/resource/aws_instance'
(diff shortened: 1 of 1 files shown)


jkeiser:
CC @tyler-ball @fnichol

tyler-ball:
Rebasing to get Travis to pass.  Otherwise, :+1:

dblessing:
Does it work to load the driver if no aws config exists? When we put our recipe together the chef run failed at compile time because aws config didn't exist yet.

tyler-ball:
I get `[2015-04-06T17:51:06-05:00] ERROR: No credentials loaded!  Do you have a ~/.aws/config?` if I don't have credentials, and it is a compile time bug.  @jkeiser is this the intended behavior?  @dblessing is your desire to have the recipe only fail if a call to AWS is attempted?  IE, we lazy load the credentials the first time an AWS call is attempted.

dblessing:
Lazy load would be great. Then the first part of recipe manage the configuration before any provisioning resources.

tyler-ball:
@dblessing Will you file a new issue on lazy loading?

dblessing:
Yep, I created one a while back. #145

Issue #159, opened 2015-03-27
Cloning resource attributes warnings at start of chef run


patrick-wright:
```
[2015-03-27T14:27:39-07:00] WARN: Cloning resource attributes for aws_network_interface[ref-eni-1] (no AWS object id) from prior resource (CHEF-3694)
[2015-03-27T14:27:39-07:00] WARN: Previous aws_network_interface[ref-eni-1] (no AWS object id): /Users/patrickwright/Chef/github/chef/chef-provisioning-aws-eni/docs/examples/aws_network_interface.rb:46:in `from_file'
[2015-03-27T14:27:39-07:00] WARN: Current  aws_network_interface[ref-eni-1] (no AWS object id): /Users/patrickwright/Chef/github/chef/chef-provisioning-aws-eni/docs/examples/aws_network_interface.rb:51:in `from_file'
[2015-03-27T14:27:39-07:00] WARN: Cloning resource attributes for aws_network_interface[ref-eni-1] (no AWS object id) from prior resource (CHEF-3694)
[2015-03-27T14:27:39-07:00] WARN: Previous aws_network_interface[ref-eni-1] (no AWS object id): /Users/patrickwright/Chef/github/chef/chef-provisioning-aws-eni/docs/examples/aws_network_interface.rb:51:in `from_file'
[2015-03-27T14:27:39-07:00] WARN: Current  aws_network_interface[ref-eni-1] (no AWS object id): /Users/patrickwright/Chef/github/chef/chef-provisioning-aws-eni/docs/examples/aws_network_interface.rb:64:in `from_file'
[2015-03-27T14:27:39-07:00] WARN: Cloning resource attributes for aws_subnet[ref-subnet-eni] (no AWS object id) from prior resource (CHEF-3694)
[2015-03-27T14:27:39-07:00] WARN: Previous aws_subnet[ref-subnet-eni] (no AWS object id): /Users/patrickwright/Chef/github/chef/chef-provisioning-aws-eni/docs/examples/aws_network_interface.rb:32:in `from_file'
[2015-03-27T14:27:39-07:00] WARN: Current  aws_subnet[ref-subnet-eni] (no AWS object id): /Users/patrickwright/Chef/github/chef/chef-provisioning-aws-eni/docs/examples/aws_network_interface.rb:72:in `from_file'
[...]

PR #166, opened 2015-03-30
Potential fix for resource cloning showing back up

tyler-ball:
Fixes https://github.com/chef/chef-provisioning-aws/issues/159

\cc @jkeiser


tyler-ball:
I need to persist both fixes so that this change works with Chef 12.1 and 12.2 (and 11.x, etc.)

jkeiser:
:+1:

PR #172, opened 2015-04-02
Add ability to ignore certain routes

dblessing:
I'm not sure if this is a great solution to the problem, but I do know it works. If nothing else it will spark some conversation about how best to handle the situation. Here's the problem:

There were 2 issues we encountered with route tables in AWS when they are managed by Chef provisioning.

1. When a VPN connection has static routes, you have the option to propagate those routes to any route table. This is necessary for traffic to flow between services on the far end of the VPN and the subnets inside AWS. If propagated routes are enabled Chef provisioning fails because it sees `vgw-abcd1234` as route targets. It tries to remove them since they aren't defined in Chef resources but it can't because they're a different kind of route. We thought maybe we could just add those routes to Chef explicitly but the API doesn't allow routes with target of `vgw-abcd1234`. It is best if we can just set these type of routes to be ignored.
2. Some things require more dynamic routes. For example, if NAT instances are set up in an HA configuration and one AZ's NAT instance dies it's peer will dynamically modify the route tables to point at a different instance or network interface. In this case, we need the ability to exclude `/^eni-/` or `/^i-/` so `i-abcd1234` and `eni-abcd1234` are ignored. 

With the changes in this PR a user can specify a route table as follows:
```
  aws_route_table 'my_route_table' do
    vpc 'my_vpc'
    routes 0.0.0.0/0 => :internet_gateway
    ignore_route_targets ['^vgw-','^i-']
  end
```

I hope these two scenarios make sense.

### DIFF
--- a/lib/chef/provider/aws_route_table.rb
+++ b/lib/chef/provider/aws_route_table.rb
@@ -7,7 +7,7 @@ class Chef::Provider::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSProvider
     route_table = super
 
     if !new_resource.routes.nil?
-      update_routes(vpc, route_table)
+      update_routes(vpc, route_table, new_resource.ignore_route_targets)
     end
   end
 
@@ -53,12 +53,13 @@ class Chef::Provider::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSProvider
 
   attr_accessor :vpc
 
-  def update_routes(vpc, route_table)
+  def update_routes(vpc, route_table, ignore_route_targets = [])
     # Collect current routes
     current_routes = {}
     route_table.routes.each do |route|
       # Ignore the automatic local route
       next if route.target.id == 'local'
+      next if ignore_route_targets.find { |target| route.target.id.match(/#{target}/) }
       current_routes[route.destination_cidr_block] = route
     end
 
--- a/lib/chef/resource/aws_route_table.rb
+++ b/lib/chef/resource/aws_route_table.rb
@@ -58,6 +58,26 @@ class Chef::Resource::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSResource
   #
   attribute :routes, kind_of: Hash
 
+  #
+  # Regex to ignore one or more route targets.
+  #
+  # This is helpful when configuring HA NAT instances. If a NAT instance fails
+  # a auto-scaling group may launch a new NAT instance and update the route
+  # table accordingly. Chef provisioning should not attempt to change or remove
+  # this route.
+  #
+  # This attribute is specified as a regex since the full ID of the
+  # instance/network interface is not known ahead of time. In most cases the
+  # NAT instance route will point at a network interface attached to the NAT
+  # instance. The ID prefix for network interfaces is 'eni'. The following
+  # example shows how to ignore network interface routes.
+  #
+  # ```ruby
+  # ignore_route_targets ['^eni-']
+  # ```
+  attribute :ignore_route_targets, kind_of: [ String, Array ], default: [],
+            coerce: proc { |v| [v].flatten }
+
   attribute :route_table_id, kind_of: String, aws_id_attribute: true, lazy_default: proc {
     name =~ /^rtb-[a-f0-9]{8}$/ ? name : nil
   }
--- a/spec/integration/aws_route_table_spec.rb
+++ b/spec/integration/aws_route_table_spec.rb
@@ -35,6 +35,38 @@ describe Chef::Resource::AwsRouteTable do
           ]
         ).and be_idempotent
       end
+
+      it "ignores routes whose target matches ignore_route_targets" do
+        expect_recipe {
+            aws_subnet 'test_subnet' do
+              vpc 'test_vpc'
+            end
+
+            eni = aws_network_interface 'test_network_interface' do
+              subnet 'test_subnet'
+            end
+
+            aws_route_table 'test_route_table' do
+              vpc 'test_vpc'
+              routes(
+                '0.0.0.0/0' => :internet_gateway,
+                '172.31.0.0/16' => eni
+              )
+            end
+  
+            aws_route_table 'test_route_table' do
+              vpc 'test_vpc'
+              routes '0.0.0.0/0' => :internet_gateway
+              ignore_route_targets ['^eni-']
+            end
+          }.to create_an_aws_route_table('test_route_table',
+            routes: [
+              { destination_cidr_block: '10.0.0.0/24', 'target.id' => 'local', state: :active },
+              { destination_cidr_block: '172.31.0.0/16' },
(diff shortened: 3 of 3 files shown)


dblessing:
I believe test failures are unrelated. I can't trigger a rebuild, though.

tyler-ball:
Hey @dblessing thanks for this PR.  I have a few questions to help understand the use case better:

1. It seems like in your first case, a better solution would be to add support for virtual private gateways or customer gateways as route targets and sources.  This way you could write something like `10.0.0.1/24 => 'vgw-abcd1234'` or `10.0.0.1/24 => vpg_resource`.  Is that how you want to structure the routes?
1. For your second case it seems something like `ignore_route_targets` would be required since AWS will automatically adjust the routes.  If they do this, will they leave the updated routes forever?  Or only until the original NAT instance comes back?

dblessing:
@tyler-ball 

1. I agreed with that initially but the API doesn't support VPG as route targets. Only internet gateways, instances and network interfaces can be set as route targets directly. That's likely because routes from the gateway are handled via the 'propagate routes' setting. 
1. Yes, the route will stay there unless another entity acts on it (like an HA NAT script would/should).

tyler-ball:
Thats what I guessed.  Based upon those answers, I think your current approach is alright, but it doesn't fit well with the resource attribute pattern.  Generally attributes are positively enforcing - you tell a `file` resource what mode it should have, not what modes it _shouldn't_ have.

If we accepted #151 would all the routes you're trying to exclude have come from the virtual private gateway?  If that is true, I don't think we would a new `ignore_route_targets` attribute.  We could simply ignore routes that are managed by the virtual private gateway.  Does that make sense, and would that work?

dblessing:
I see what you're saying. If you accept #151 that will cover use-case 1 (only if some additional logic is included to automatically ignore those routes when enabled). 

Use-case 2 is still unsolved without this or a similar solution. There is no possibility of having highly-available NAT instances without some ability to ignore routes. Chef provisioning can provision the route table with the network interface associated with the NAT instance Chef provisioning builds *at that time*. However, any time in the future that instance could die, be autoscaled, and the HA NAT script would come along as update the route dynamically. If Chef doesn't have a way to ignore this it will come along and see a 'problem' and try to set it back to the original interface. Any other ideas how we can accomplish this without a negative attribute?

tyler-ball:
We are both in agreement that as part of #151, routes that come from the attribute `vgw_route_propagation` should not be affected by the attribute `routes`, and vice versa.  

The proposed solution here is that any other routes not covered by those two attributes are managed UNLESS they match the `ignore_route_targets` attribute.  No, I don't have any better ideas to manage routes that are automatically updated by AWS, and I think your solution is fine.  I would like to get @jkeiser's opinion on this, but otherwise I'm :+1:

tyler-ball:
Before this PR can be accepted I would like to see some tests added to https://github.com/chef/chef-provisioning-aws/blob/master/spec/integration/aws_route_table_spec.rb.

lib/chef/resource/aws_route_table.rb
```diff
@@ -58,6 +58,8 @@ class Chef::Resource::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSResource
   #
   attribute :routes, kind_of: Hash
 
+  attribute :ignore_route_targets, kind_of: Array, default: []
```
tyler-ball: Can you add some RDoc here that explains what users are expected to provide in their cookbooks, and how that will change their route table management?

dblessing:
See https://github.com/chef/chef-provisioning-aws/pull/151#issuecomment-101372684 - running integration tests locally is not all that feasible for me.

dblessing:
@tyler-ball Can you explain how I might test this? I looked at the integration tests but can't think of a way to create a route table with a route that I can then go back and ignore and ensure it's still there.

tyler-ball:
I'm thinking of something like

```patch
diff --git a/spec/integration/aws_route_table_spec.rb b/spec/integration/aws_route_table_spec.rb
index ef5a8fe..748bafb 100644
--- a/spec/integration/aws_route_table_spec.rb
+++ b/spec/integration/aws_route_table_spec.rb
@@ -35,6 +35,26 @@ describe Chef::Resource::AwsRouteTable do
           ]
         ).and be_idempotent
       end
+
+      it "ignores routes whose target matches ignore_route_targets" do
+        expect_recipe {
+          aws_route_table 'test_route_table' do
+            vpc 'test_vpc'
+            routes '0.0.0.0/0' => :internet_gateway
+          end
+
+          aws_route_table 'test_route_table' do
+            vpc 'test_vpc'
+            routes '0.0.0.1/0' => :internet_gateway
+            ignore_route_targets '^ig-'
+          end
+        }.to create_an_aws_route_table('test_route_table',
+          routes: [
+            { destination_cidr_block: '10.0.0.0/24', 'target.id' => 'local', state: :active },
+            { destination_cidr_block: '0.0.0.0/0', 'target.id' => test_vpc.aws_object.internet_gateway.id, state: :active }
+          ]
+        ).and be_idempotent
+      end
     end
   end
 end

```

lib/chef/provider/aws_route_table.rb
```diff
@@ -7,7 +7,7 @@ def action_create
     route_table = super
 
     if !new_resource.routes.nil?
-      update_routes(vpc, route_table)
+      update_routes(vpc, route_table, [new_resource.ignore_route_targets].flatten)
```
tyler-ball: As I referenced in your [other review](https://github.com/chef/chef-provisioning-aws/pull/151/files#r30183218) you can coerce the `ignore_route_targets` attribute in the resource so you always know it will be an array in here

spec/integration/aws_route_table_spec.rb
```diff
@@ -35,6 +35,38 @@
...
+  
+            aws_route_table 'test_route_table' do
+              vpc 'test_vpc'
+              routes '0.0.0.0/0' => :internet_gateway
+              ignore_route_targets ['^eni-']
+            end
+          }.to create_an_aws_route_table('test_route_table',
+            routes: [
+              { destination_cidr_block: '10.0.0.0/24', 'target.id' => 'local', state: :active },
+              { destination_cidr_block: '172.31.0.0/16' },
```
dblessing: This passes. I'm not checking the target because I wasn't sure if/how I could get the network interface ID. Any tips?

dblessing:
I went with the network interface approach here because the internet gateway method from the snippet you posted won't work. If we set an ignore on '^igw-' for example, it won't allow any management of internet gateways so the second definition of the route table wouldn't work. Does that make sense?

tyler-ball:
@dblessing Yup, that makes total sense!  I updated the test on merging to check the network interface id - you can see what I did in https://github.com/chef/chef-provisioning-aws/commit/70195b42ea2a8e35b0c134bd0de9c8e954e83086

Thanks again for all the awesome work!

PR #175, opened 2015-04-06
Add ability to turn off source/dest check

dblessing:
Source/dest check needs to be off for NAT instances. This allows `source_dest_check: false` to be set as a `machine_options` key/value.


jkeiser:
:+1: ... let's set us up a test quick.

jkeiser:
Also, hooray for NAT!

I'm working out a test, be a minute.

jkeiser:
Rebased and merged as 41d7005

PR #181, opened 2015-04-09
Updated security group examples to use the correct hash key for ports.

msonnabaum:
This one is especially tricky since it doesn't error when using "ports", but the rule gets created with a port range of 0..0.


tyler-ball:
Thanks for the catch!

PR #185, opened 2015-04-15
update aws_ebs_volume :availability_zone to exclude region setting

patrick-wright:
resources now look like:
```ruby
aws_ebs_volume 'ref-volume' do
  availability_zone 'a'
  size 1
end
```

### DIFF
--- a/lib/chef/provider/aws_ebs_volume.rb
+++ b/lib/chef/provider/aws_ebs_volume.rb
@@ -47,7 +47,7 @@ class Chef::Provider::AwsEbsVolume < Chef::Provisioning::AWSDriver::AWSProvider
 
   def update_aws_object(volume)
     if initial_options.has_key?(:availability_zone)
-      if initial_options[:availability_zone] != volume.availability_zone_name
+      if availability_zone != volume.availability_zone_name
         raise "#{new_resource}.availability_zone is #{availability_zone}, but actual volume has availability_zone_name set to #{volume.availability_zone_name}.  Cannot be modified!"
       end
     end
@@ -206,6 +206,12 @@ class Chef::Provider::AwsEbsVolume < Chef::Provisioning::AWSDriver::AWSProvider
   end
 
   def availability_zone
-    "#{region}#{new_resource.availability_zone}"
+    az = new_resource.availability_zone
+    # resource attribute has already validated the value with regex
+    if az.length > 1
+      Chef::Log.warn("availability_zone attribute should only be set to the letter designation. Attempting to use '#{az[-1]}' to correct the issue.")
+      az = az[-1]
+    end
+    "#{region}#{az}"
   end
 end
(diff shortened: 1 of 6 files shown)


patrick-wright:
@tyler-ball

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -204,4 +204,8 @@ def delete(volume)
       volume
     end
   end
+
+  def availability_zone
+    "#{region}#{new_resource.availability_zone}"
```
tyler-ball: Now that 1.0 is out, we need to support backwards compatability.  Can you add in a regex that checks if `new_resource.availability_zone` begins with `region`?  If so, it should output a WARN message about the deprecated logic and then only return the provided availability_zone.  Otherwise it should do the logic here.

lib/chef/provider/aws_ebs_volume.rb
```diff
@@ -204,4 +204,14 @@ def delete(volume)
       volume
     end
   end
+
+  def availability_zone
```
patrick-wright: @tyler-ball please review these changes

spec/integration/aws_ebs_volume_spec.rb
```diff
@@ -6,16 +6,34 @@
...
+         .and be_idempotent
+      end
+
+      it "aws_ebs_volume fails with an invalid AZ pattern", :skip => true do
+        expect_recipe {
+          aws_ebs_volume "test_volume_az0" do
+            availability_zone 'usa-abcd-1'
+            size 8
+          end
+        }.to raise_error(Chef::Exceptions::ValidationFailed, /availability_zone/)
```
patrick-wright: @tyler-ball I have not been able to expect an error.  I get an rspec error:
```
     Failure/Error: expect_recipe {
       expected Chef::Exceptions::ValidationFailed but was not given a block
```

lib/chef/provider/aws_ebs_volume.rb
tyler-ball: But they could set their region to `eu-west-1` in their `with_driver` block and write their availability_zone as `us-east-2a`.  With the current code, it would put their EBS volume in `eu-west-1a`.

If you do something like the following:
```ruby
if /^#{region}/ =~ az
  Chef::Log.warn("availability_zone attribute should only be set to the letter designation. Attempting to use '#{az[-1]}' to correct the issue.")
elsif az.length == 1
  az = "#{region}#{az[-1]}"
end
```

Then you ensure that (if provided) the full availability zone matches the current region.  If it doesn't match, amazon will get angry that they're trying to put it in an availability zone that doesn't exist in the region.

What validation are you talking about?  I couldn't find any in the provider or the resource.

spec/integration/aws_ebs_volume_spec.rb
tyler-ball: I think for this you want to use the old syntax:

```ruby
expect {
  converge do
    aws_ebs_volume "test_volume_az0" do
      availability_zone 'usa-abcd-1'
      size 8
    end
  end
}.to raise_error
```

lib/chef/provider/aws_ebs_volume.rb
patrick-wright: True, but this maintains the previous functionality, and we didn't validate for that before letting EC2 catch that specific exception.  You're code looks nicer anyhow, so I'll make the change.

PR #186, opened 2015-04-15
add destroy_an_aws_object

patrick-wright:
the test spec may cause a merge conflict, so w/e

@tyler-ball @curiositycasualty

### DIFF
--- a/spec/integration/aws_ebs_volume_spec.rb
+++ b/spec/integration/aws_ebs_volume_spec.rb
@@ -17,6 +17,23 @@ describe Chef::Resource::AwsEbsVolume do
         ).and be_idempotent
       end
 
+      describe 'action :delete' do
+        with_converge {
+          aws_ebs_volume "test_volume" do
+            availability_zone 'a'
+            size 8
+          end
+        }
+        it "deletes the ebs volume" do
+          expect_recipe {
+            aws_ebs_volume "test_volume" do
+              action :destroy
+            end
+          }.to destroy_an_aws_ebs_volume('test_volume')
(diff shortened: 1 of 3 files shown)


spec/integration/aws_ebs_volume_spec.rb
```diff
@@ -9,11 +9,18 @@
...
           aws_ebs_volume "test_volume" do
-            availability_zone 'eu-west-1a'
+            availability_zone "#{driver.aws_config.region}a"
             size 8
           end
         }.to create_an_aws_ebs_volume('test_volume')
          .and be_idempotent
+
+        expect_recipe {
+          aws_ebs_volume "test_volume" do
```
tyler-ball: This test shouldn't be passing right now.  You should need something like:
```ruby
aws_ebs_volume "test_volume" do
  availability_zone "a"
end
aws_ebs_volume "test_volume" do
  action :destroy
end
```

Otherwise there is no resource to destroy and the matcher should be failing

patrick-wright: @tyler-ball take a closer look :)

tyler-ball: Oh, its all in the same test :facepalm:.  Okay, I think that makes sense.

patrick-wright: If its any assurance, I ran the same test by changing line
```ruby
 }.to destroy_an_aws_ebs_volume('test_volume_ABC')
```
before I pushed changes and the test failed in the way I expected.

PR #187, opened 2015-04-16
Updating destroy to remove instances and images on non-purge destroy

tyler-ball:
\cc @jkeiser @fnichol


jkeiser:
:+1:  Can you leave an issue out there to wait for termination of other types as well?  (Unless they already are.)

tyler-ball:
Sure, I'll throw an issue up there

tyler-ball:
Actually I'm not sure any other resources need a wait.  Most other things immediately disappear when you call :destroy

Issue #189, opened 2015-04-22
machine :destroy RuntimeError when aws_instance resource has been used in the recipe


patrick-wright:
### recipe
```ruby
require 'chef/provisioning/aws_driver'

with_driver 'aws::us-west-2'

machine 'ref-machine-1' do
  action :allocate
end

aws_instance 'ref-machine-1' do
  aws_tags :marco => 'polo', :happyhappy => 'joyjoy'
end

aws_instance 'ref-machine-1' do
  aws_tags :othercustomtags => 'byebye'
end

aws_instance 'ref-machine-1' do
  aws_tags :Name => 'new-name'
end

machine 'ref-machine-1' do
  action :destroy
end

# Same error using:
# aws_instance 'ref-machine-1' do
#   action :destroy
# end
```

### client
```
================================================================================
    Error executing action `destroy` on resource 'machine[ref-machine-1]'
    ================================================================================

    RuntimeError
    ------------
    Switching a machine's driver from  to aws::us-west-2 is not currently supported!  Use machine :destroy and then re-create the machine on the new driver.

    Resource Declaration:
    ---------------------
    # In /Users/patrickwright/Chef/github/chef/chef-provisioning-aws/docs/examples/aws_tags.rb

     21: machine 'ref-machine-1' do
     22:   action :destroy
     23: end

    Compiled Resource:
    ------------------
    # Declared in /Users/patrickwright/Chef/github/chef/chef-provisioning-aws/docs/examples/aws_tags.rb:21:in `from_file'

    machine("ref-machine-1") do
      action [:destroy]
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      chef_server {:chef_server_url=>"http://localhost:8889", :options=>{:client_name=>"remwright01", :signing_key_filename=>nil}}
      driver "aws::us-west-2"
      declared_type :machine
      cookbook_name "@recipe_files"
      recipe_name "/Users/patrickwright/Chef/github/chef/chef-provisioning-aws/docs/examples/aws_tags.rb"
    end
```

### stacktrace
```
RuntimeError: machine[ref-machine-1] (@recipe_files::/Users/patrickwright/Chef/github/chef/chef-provisioning-aws/docs/examples/aws_tags.rb line 21) had an error: RuntimeError: Switching a machine's driver from  to aws::us-west-2 is not currently supported!  Use machine :destroy and then re-create the machine on the new driver.
[...]

PR #190, opened 2015-04-22
Aws Tags

patrick-wright:
Support tagging for AWS resources which are "taggable".

### DIFF
--- a/docs/examples/aws_tags.rb
+++ b/docs/examples/aws_tags.rb
@@ -1,0 +1,71 @@
+require 'chef/provisioning/aws_driver'
+
+with_driver 'aws::us-west-2'
+
+machine 'ref-machine-1' do
+  action :allocate
+end
+
+machine 'ref-machine-1' do
+  machine_options aws_tags: {:marco => 'polo', :happyhappy => 'joyjoy'}
+  converge false
+end
+
+machine 'ref-machine-1' do
+  machine_options aws_tags: {:othercustomtags => 'byebye'}
+  converge false
+end
+
+machine 'ref-machine-1' do
+  machine_options aws_tags: {:Name => 'new-name'}
+  converge false
+end
+
+machine 'ref-machine-1' do
+  action :destroy
+end
+
+machine_batch "ref-batch" do
+  machine 'ref-machine-2' do
+    machine_options aws_tags: {:marco => 'polo', :happyhappy => 'joyjoy'}
+    converge false
+  end
+  machine 'ref-machine-3' do
+    machine_options aws_tags: {:othercustomtags => 'byebye'}
+    converge false
+  end
+end
+
+load_balancer 'ref-elb' do
+  load_balancer_options availability_zones: ['us-west-1a', 'us-west-1b']
+end
+
+load_balancer 'ref-elb' do
+  load_balancer_options aws_tags: {:marco => 'polo', :happyhappy => 'joyjoy'}
+end
+
+load_balancer 'ref-elb' do
+  load_balancer_options aws_tags: {:othercustomtags => 'byebye'}
+end
+
+load_balancer 'ref-elb' do
+  action :destroy
+end
+
+machine_batch "ref-batch" do
+  action :destroy
+end
+
+machine_image "ref-image" do
+  image_options aws_tags: {:marco => 'polo', :happyhappy => 'joyjoy'}
+end
+
+# There is a bug where machine_images do not update - so we cannot update
+# the tags on it
+machine_image "ref-image" do
+  image_options aws_tags: {:othercustomtags => 'byebye'}
+end
+
+machine_image "ref-image" do
+  action :destroy
+end
--- a/lib/chef/provisioning/aws_driver/aws_provider.rb
+++ b/lib/chef/provisioning/aws_driver/aws_provider.rb
@@ -121,6 +121,8 @@ class AWSProvider < Chef::Provider::LWRPBase
       aws_object = create_aws_object
     end
 
+    converge_tags(aws_object)
+
     #
     # Associate the managed entry with the AWS object
     #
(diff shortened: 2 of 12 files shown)


patrick-wright:
@tyler-ball this is ready for a review

patrick-wright:
@chef/ociv work re: aws tagging

tyler-ball:
The tag example is still failing like https://github.com/chef/chef-provisioning-aws/issues/189 correct?

docs/examples/aws_tags.rb
```diff
@@ -0,0 +1,23 @@
+require 'chef/provisioning/aws_driver'
+
+with_driver 'aws::us-west-2'
+
+machine 'ref-machine-1' do
+  action :allocate
+end
+
+aws_instance 'ref-machine-1' do
```
tyler-ball: Hmm - I totally didn't think about the fact that most people are going to use `machine` and that doesn't support tag attributes.  This solution seems less than intuitive for people.  And I don't want to add a `tag` attribute to `machine` because that would require a new release of chef-provisioning...  And whatever solution we use will have to work for machine, machine_batch and load_balancer.

Its not awesome, but I think we should read a `tags` field in the machine_options like so:

```ruby
machine 'test' do
  machine_options bootstrap_options: {
    aws_tags: {'key' => 'value'}
  }
end
machine_batch 'test-batch' do
  machine_options bootstrap_options: {
    aws_tags: {'key' => 'value'}
  }
  machine 'test1'
  machine 'test2'
end

load_balancer 'test_lb' do
  load_balancer_options aws_tags: {'key' => 'value'}
end
```

@jkeiser what do you think of this?  Its a weird dual-user-interface problem, but I don't want to add a base `provider_tags` attribute to these resources in chef-provisioning because that implies all providers support tags.

Long term, I want to fix this by having `machine` in a recipe lookup the correct `AwsInstance` resource using the new resource mapping introduced in Chef 12.  Then
```ruby
machine 'ref-machine-1` do
  aws_tags :marco => 'polo'
end
```
would correctly instantiate a `AwsInstance` resource and apply the `aws_tags` to it.

lib/chef/provisioning/aws_driver/aws_provider.rb
```diff
@@ -221,6 +223,29 @@ def destroy_aws_object(obj)
     raise NotImplementedError, :destroy_aws_object
   end
 
+  # TODO documentation and tests
```
tyler-ball: Can you add some yarddoc here?

patrick-wright:
Yes, that's how I discovered it.  The aws tag spec that tags an aws_instance is terminated properly.  I was hoping it was some issue with the recipe I am overlooking.

spec/integration/aws_tagged_items_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        }.to have_aws_ebs_volume_tags('test_volume_a',
+                                      { 'Name' => 'test_volume_b',
+                                        'project' => 'X'
+                                      })
+      end
+
+      it "aws_ebs_volume 'test_volume' created with new Name tag" do
+        expect_recipe {
+          aws_ebs_volume "test_volume_2" do
+            aws_tags :Name => 'test_volume_new'
```
tyler-ball: Can you make the key here `'Name'` instead of `:Name` so we can test that works?

docs/examples/aws_tags.rb
patrick-wright: I'm +1 on using *_options for instances and load balancers over how it is currently.

lib/chef/provisioning/aws_driver/aws_provider.rb
patrick-wright: will do

tyler-ball:
The specs and the matcher look great!  There is only a few missing specs I would like to see:
```ruby
aws_ebs_volume 'ref' do
  aws_tags {'aaa' => 'a'}
end
aws_ebs_volume 'ref' do
  aws_tags {'bbb' => 'b'}
end
```
should only have the `'bbb'` tag on it after convergence.  This shows that deleting a tag by not putting it in the tag list works.

The other spec should test that when the `aws_tags` is missing we don't modify any of the current tags:
```ruby
aws_ebs_volume 'ref' do
  aws_tags {'aaa' => 'a'}
end
aws_ebs_volume 'ref'
```
should still contain the tags `'aaa'`

docs/examples/aws_tags.rb
patrick-wright: I'd also like to add that even though tagging machines and load balancers isn't ideal with this branch, I would press that we still merge these changes pending a good review.  This is a feature that is much needed and we can improve it in a new branch.

patrick-wright:
Good all on the tests.  The first case I was handling, but the second case uncovered a bug where the tags were being wiped out because the default aws_tags attribute is {}.  Since we may want that functionality I removed the default {} and handle the nil attribute in #converge_tags.

spec/integration/aws_tagged_items_spec.rb
patrick-wright: done

lib/chef/resource/aws_eip_address.rb
```diff
@@ -6,6 +6,9 @@ class Chef::Resource::AwsEipAddress < Chef::Provisioning::AWSDriver::AWSResource
 
   attribute :name, kind_of: String, name_attribute: true
 
+  # AWS really sucks - every other AWSResourceWithEntry accepts tags EXCEPT this one
```
patrick-wright: I'll alter this comment

spec/aws_support/matchers/have_aws_object_tags.rb
```diff
@@ -0,0 +1,46 @@
...
+        resource = resource_class.new(name, nil)
+        resource.driver example.driver
+        resource.managed_entry_store Chef::Provisioning.chef_managed_entry_store
+        aws_object = resource.aws_object
+
+        # Check existence
+        if aws_object.nil?
+          differences << "#{resource_name}[#{name}] did not exist!"
+        else
+          differences += match_hashes_failure_messages(expected_tags, aws_object.tags.to_h, "#{resource_name}[#{name}]")
```
patrick-wright: I should either come up with some valid differences or just remove this entirely

tyler-ball:
@patrick-wright you should be able to rebase against master and have your ref work!

tyler-ball:
:+1: Great job!

danieljimenez:
It doesn't seem to work on security groups when you use 'add_machine_options'

patrick-wright:
@danieljimenez you need to use the aws_tags attribute with the resource.
```ruby
aws_security_group 'blah' do
  aws_tags :tag_name => 'value'
 # other attrs
end
```
the readme has some additional details: https://github.com/chef/chef-provisioning-aws#aws-resources

danieljimenez:
@patrick-wright I did this:

```
add_machine_options aws_tags: { 'my-tag' => 'tag' }
... build machines ...
aws_security_group 'sg-12345'
```

patrick-wright:
@danieljimenez good find.
@tyler-ball check this out: https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/resource/aws_security_group.rb#L1-L5

Looks like AwsSecurityGroup resource is requiring  _with_entry, but is subclassing AWSResource.  I'll try updating this locally, but do you see an immediate issue changing that to AWSResourceWithEntry?

tyler-ball:
@patrick-wright Ah!  I updated security groups to subclass `AWSResource` _instead of_ `AWSResourceWithEntry` in https://github.com/chef/chef-provisioning-aws/pull/194.  I did this because the name of an aws security group is specific to a VPC, so we don't need to store a mapping entry.

It looks like this exposed an issue with tags!  I'll file a bug.

tyler-ball:
Tracking this new issue in https://github.com/chef/chef-provisioning-aws/issues/204

PR #191, opened 2015-04-23
Updating to use the new *spec.driver_url syntax exposed in chef-provisioning 1.0

tyler-ball:
Fixes https://github.com/chef/chef-provisioning-aws/issues/189
\cc @patrick-wright @jkeiser

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -90,10 +90,10 @@ module AWSDriver
           actual_elb = elb.load_balancers.create(lb_spec.name, lb_options)
 
           lb_spec.reference = {
-            'driver_url' => driver_url,
             'driver_version' => Chef::Provisioning::AWSDriver::VERSION,
             'allocated_at' => Time.now.utc.to_s,
           }
+          lb_spec.driver_url = driver_url
         end
       else
         # Header gets printed the first time we make an update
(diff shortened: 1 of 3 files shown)


lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -603,8 +603,8 @@ def load_balancer_for(lb_spec)
 
     def instance_for(machine_spec)
       if machine_spec.reference
-        if machine_spec.reference['driver_url'] != driver_url
```
tyler-ball: The underlying issue in #189 is that `driver_url` is no longer stored in the reference by machine, because it is using the new API exposed in chef-provisioning 1.0

patrick-wright:
I don't see any red flags :+1:

tyler-ball:
To test this in a unit test we would need A LOT more scaffolding.  I tried out the `ref_full` test and that worked for me, so I would call that testing good enough.

jkeiser:
:+1:

We can probably do something to write an integration test, but it'd take some work.  I can imagine it starting with:

```ruby
context 'When there is an new-style node pointed at an instance ID" do
  machine 'blah' do
    action :allocate
  end

  before :each do
    converge do
      node = Chef::Node.load('blah')
      node.normal['chef_provisioning'].delete('driver_url')
      node.save
    end
  end

  it "machine 'blah' { action :destroy } works" do
    expect_recipe {
      machine 'blah' do
        action :destroy
      end
    }.to be_updated
  end
end
```

tyler-ball:
I don't know what I was thinking - I was able to add a test which duplicated the original issue.  I validated that the test failed against the old branch.

PR #192, opened 2015-04-28
Updating ref files to run correctly

tyler-ball:
There are a few failures with the current ref files:

1. Creating objects in the EU is triggering timeouts trying to ssh into the machines, even they are created successfully.  Updating to faster instances in a local AZ.
1. Machine images are not correctly waiting for their template machines to become terminated before shutting themselves down
1. You cannot delete the main route table or update the main route table except through creating a new main route table.  This means only a `:purge` on the VPC will delete the default route table created and delete the currently assigned route table.

### DIFF
--- a/docs/examples/ref_destroy.rb
+++ b/docs/examples/ref_destroy.rb
@@ -1,6 +1,6 @@
 require 'chef/provisioning/aws_driver'
 
-with_driver 'aws::eu-west-1'
+with_driver 'aws::us-west-1'
 
 aws_sqs_queue 'ref-sqs-queue' do
   action :destroy
@@ -69,6 +69,19 @@ end
 
 aws_key_pair 'ref-key-pair' do
   action :destroy
+end
+
+# You cannot delete the main route table, or delete a VPC which
+# has non-main route tables attached.  So we first need to restore
+# the 'default' route tabled created during the `create_vpc`
+# call as the main route table.  Then we can delete the
+# 'ref-main-route-table' (because it is no longer main)
+# and finally delete the VPC (which deletes the main route table)
+aws_vpc 'ref-vpc' do
+  main_route_table lazy {
+    self.aws_object.route_tables.select {|r| !r.main?}.first
+  }
+  only_if { !self.aws_object.nil? }
 end
 
 aws_route_table 'ref-main-route-table' do
--- a/docs/examples/ref_full.rb
+++ b/docs/examples/ref_full.rb
@@ -1,14 +1,14 @@
 require 'chef/provisioning/aws_driver'
 
-with_driver 'aws::eu-west-1'
+with_driver 'aws::us-west-1'
 
 #
 # This recipe sets every single value of every single object
 #
 
-aws_vpc 'ref-vpc' do
-  action :purge
-end
+# aws_vpc 'ref-vpc' do
+#   action :purge
+# end
 
 aws_dhcp_options 'ref-dhcp-options' do
   domain_name          'example.com'
@@ -37,15 +37,14 @@ aws_vpc 'ref-vpc' do
   main_route_table 'ref-main-route-table'
 end
 
-
 aws_key_pair 'ref-key-pair' do
   private_key_options({
-    :format => :der,
-    :type => :dsa
+    :format => :pem,
+    :type => :rsa,
+    :regenerate_if_different => true
   })
   allow_overwrite true
 end
-
 
 aws_security_group 'ref-sg1' do
   vpc 'ref-vpc'
@@ -55,7 +54,6 @@ aws_security_group 'ref-sg1' do
   ]
 end
 
-
 aws_route_table 'ref-public' do
   vpc 'ref-vpc'
   routes '0.0.0.0/0' => :internet_gateway
@@ -64,7 +62,7 @@ end
 aws_subnet 'ref-subnet' do
   vpc 'ref-vpc'
   cidr_block '10.0.0.0/24'
-  availability_zone 'eu-west-1a'
+  availability_zone 'us-west-1a'
   map_public_ip_on_launch true
   route_table 'ref-public'
 end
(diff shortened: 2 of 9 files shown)


tyler-ball:
Hey @jkeiser - I discovered some stuff!  You cannot delete a VPC if it has ANY non-main route tables attached.  You also cannot update a VPC or existing route table to make it the main route table except through the web console.

My idea is to start managing the default created route table completely separately from one assigned from another resource.  This would mean the user cannot both specify `main_routes` and `main_route_table`.  If the main route on a default route table is updated, we will update the default route table.  If the main route table in the resource is updated, we will delete the default route table and attach the provided resource.  Then if the main routes are again supplied, we will detach the resource route table and create a new default route table (attaching it as the main).

What do you think?

lib/chef/provider/aws_vpc.rb
```diff
@@ -160,37 +161,42 @@ def update_internet_gateway(vpc)
...
 
-  def update_main_routes(vpc, main_route_table)
-    main_route_table ||= vpc.route_tables.main_route_table
-    aws_route_table main_route_table do
-      vpc vpc
-      routes new_resource.main_routes
+  # @param main_routes [Hash]
+  def update_main_routes(vpc, main_routes)
+    require 'pry'; binding.pry
+    ::Chef::Provisioning.inline_resource(action_handler) do
```
tyler-ball: @jkeiser this is what I'm having trouble with.  When I try to do this, I'm missing the `desired_driver` in the `aws_provider`.

docs/examples/ref_destroy.rb
```diff
@@ -71,12 +71,24 @@
...
-aws_route_table 'ref-main-route-table' do
-  action :destroy
+# TODO fix this so it creates a new default route table and detaches an existing one
+aws_vpc 'ref-vpc' do
+  main_routes '0.0.0.0/0' => :internet_gateway
 end
 
+# The only call that changes the MAIN route table is the
+# 'create route table' call - there is no update except
+# through the console, and you cannot delete the main
```
jkeiser: Did you find a way to improve the error message?

docs/examples/ref_destroy.rb
```diff
@@ -71,12 +71,24 @@
...
+# all non-main route tables, and then the delete
+# vpc call automatically deletes the main route table
+
+# aws_route_table 'ref-main-route-table' do
+#   action :destroy
+# end
+
 aws_vpc 'ref-vpc' do
-  action :destroy
+  action :purge
```
jkeiser: Do we know why `purge` manages to destroy the main route table while `destroy` does not?  It seems like `destroy` ought to have been sufficient here.

lib/chef/provider/aws_route_table.rb
```diff
@@ -31,14 +31,20 @@ def update_aws_object(route_table)
...
   end
 
   def destroy_aws_object(route_table)
-    converge_by "delete route table #{new_resource.name} (#{route_table.id}) in #{region}" do
-      route_table.delete
+    converge_by "delete route table #{new_resource.to_s} in #{region}" do
+      begin
+        route_table.delete
+      rescue AWS::EC2::Errors::DependencyViolation
+        raise "#{new_resource.to_s} could not be deleted because it is the main route table for #{route_table.vpc.id}"
```
tyler-ball: TODO - test if this is the same error raised when trying to delete a route table in use by a subnet.  If so, change the error to say it is either the main route table for a VPC or is used by a subnet.

tyler-ball:
TODO - can we use http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ReplaceRouteTableAssociation.html to replace an assoication between a VPC and the main route table?

docs/examples/ref_destroy.rb
```diff
@@ -71,12 +71,24 @@
   action :destroy
 end
 
-aws_route_table 'ref-main-route-table' do
-  action :destroy
+# TODO fix this so it creates a new default route table and detaches an existing one
+aws_vpc 'ref-vpc' do
```
tyler-ball:
```ruby
ruby_block 'delete implicetly created route table' do
  block do
   implicit_routes = vpc.routes.select |r| {r.tags.include 'created_by_chef'}
   implicit_routes.each do |r| {r.delete}
  end
end
```

lib/chef/provider/aws_vpc.rb
```diff
@@ -18,12 +19,12 @@ def action_create
 
     # Replace the main route table for the VPC
     if !new_resource.main_route_table.nil?
-      main_route_table = update_main_route_table(vpc)
+      update_main_route_table(vpc, new_resource.main_route_table)
```
tyler-ball: TODO - don't do this, update whatever the main route table is, don't create an implicit route table

tyler-ball:
\cc @jkeiser This is now ready for review but depends on https://github.com/chef/chef-provisioning/pull/337

docs/examples/ref_full.rb
```diff
@@ -78,18 +76,28 @@
 end
 
 machine_image 'ref-machine_image3' do
-  machine_options bootstrap_options: { subnet_id: 'ref-subnet', security_group_ids: 'ref-sg1', image_id: 'ref-machine_image1' }
+  machine_options bootstrap_options: {
+    # for some reason, sshing into this host takes 20+ seconds with these enabled
+    #subnet_id: 'ref-subnet',
+    #security_group_ids: 'ref-sg1',
```
tyler-ball: TODO: uncomment security_group_ids

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -359,22 +359,12 @@ def ready_image(action_handler, image_spec, image_options)
...
-        end
-        if snapshots.any?
-          action_handler.perform_action "Deleting image #{image_spec.name} snapshots" do
-            snapshots.each do |snap|
-              snap.delete
-            end
-          end
+      # TODO the driver should automatically be set by `inline_resource`
+      d = self
+      Provisioning.inline_resource(action_handler) do
```
tyler-ball: TODO try propogating the driver, the event queue, chef_provisioning, etc.

We could also call `with_driver`, `with_chef_server`, etc. at the beginning of the recipe.

Make `Provisioning.inline_resource` accept driver as an optional parameter.

PR #197, opened 2015-05-06
Query Ubuntu for current AMI

### LINKED ISSUE #196: Ask Ubuntu for the latest Ubuntu image as the default image instead of hardcoding

whiteley:
* find AMIs using `ubuntu_ami` gem
* fixes #196 

Does this seem like a reasonable dependency to add? Any changes you'd like to see?

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -21,7 +21,7 @@ require 'chef/provisioning/aws_driver/credentials'
 require 'yaml'
 require 'aws-sdk-v1'
 require 'retryable'
-
+require 'ubuntu_ami'
 
 # loads the entire aws-sdk
 AWS.eager_autoload!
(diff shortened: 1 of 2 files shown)


lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -675,31 +675,42 @@ def aws_credentials
...
+    def default_ami_release
+      'trusty'
+    end
+
+    def default_ami_root_store
+      'ebs'
+    end
+
+    def default_ami_virtualization_type
+      'paravirtual'
```
tyler-ball: Did you verify that these defaults are the current values for the existing AMIs?

tyler-ball:
I like it!  I have one question, but otherwise I'm :+1:

lib/chef/provisioning/aws_driver/driver.rb
whiteley: I did but they changed ... When I made this PR, they were `trusty/ebs/paravirtual` but the current ones are `utopic/ebs/hvm`. I would think they would be the LTS version or the current version. Should I update it to `utopic` or `vivid`?

tyler-ball: Lets do the latest, vivid.  We're trying to keep these defaults current.  I honestly don't know the difference between hvm and paravirtual, but in the interest of not changing too much backwards compatability I would say hvm.  How do you feel about that?

whiteley: I think both are fine choices as defaults. New instance types seem to be HVM only as are the burstable instances people might use more with the free tier.

whiteley:
Updated `default_ami_release` and `default_ami_virtualization_type`

tyler-ball:
@whiteley have you ever ran the [integration tests](https://github.com/whiteley/chef-provisioning-aws/blob/master/spec/integration/machine_spec.rb)?  If not, will you run them locally with something like `AWS_TEST_DRIVER=aws::eu-west-1 rspec spec/integration/aws_machine_spec.rb -t super_slow`?  This will use your local `~/.aws/credentials` or environment variables to auth with AWS.

We need to get these integration tests running via Travis, we just haven't yet.

whiteley:
Looks like I only ran the unit tests before, I just ran them and I fixed one issue. I'm getting `Chef::Exceptions::PrivateKeyMissing`, do I need to provide a `client_key` someplace? I think the other failures I'm seeing are because the account I'm on is an old one without VPC by default but I'm not sure.

tyler-ball:
@whiteley Hmm, no you shouldn't...  It should create a key by default in `~/.chef/keys/chef_default`.  Would you mind filing a new issue showing the fill stacktrace you get when you try to run the tests?

I'm going to go ahead and merge this because I don't want our broken tests to stop your PR.  Thanks again for this work!

PR #199, opened 2015-05-12
Support elb attributes

dblessing:
Add support for load balancer attributes (Adds ability to enable cross-zone load balancing, idle timeouts, etc.) See http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/ELB/Client/V20120601.html#modify_load_balancer_attributes-instance_method for more information. The implementation seems a little wordy but it is just re-using the same terms that the AWS SDK uses. This leaves us with a very verbose implementation:

```
load_balancer 'ref-load-balancer' do
  machines [ 'ref-machine2' ]
  load_balancer_options(
    attributes: {
      cross_zone_load_balancing: {
        enabled: true
      }
    }
  )
end
```

But it works :)

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -1,4 +1,5 @@
 require 'chef/mixin/shell_out'
+require 'chef/mixin/deep_merge'
 require 'chef/provisioning/driver'
 require 'chef/provisioning/convergence_strategy/install_cached'
 require 'chef/provisioning/convergence_strategy/install_sh'
@@ -33,6 +34,7 @@ module AWSDriver
   class Driver < Chef::Provisioning::Driver
 
     include Chef::Mixin::ShellOut
+    include Chef::Mixin::DeepMerge
 
     attr_reader :aws_config
 
@@ -69,6 +71,9 @@ module AWSDriver
     # Load balancer methods
     def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
       lb_options = AWSResource.lookup_options(lb_options || {}, managed_entry_store: lb_spec.managed_entry_store, driver: self)
+      # We delete the attributes here because they are not valid in the create call
+      # and must be applied afterward
+      lb_attributes = lb_options.delete(:attributes)
 
       old_elb = nil
       actual_elb = load_balancer_for(lb_spec)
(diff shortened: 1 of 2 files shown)


dblessing:
Ready for review. Thanks :smile:

tyler-ball:
@dblessing How about just `attributes` instead of `load_balancer_attributes`?

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -67,6 +67,7 @@ def self.canonicalize_url(driver_url, config)
     # Load balancer methods
     def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
       lb_options = AWSResource.lookup_options(lb_options || {}, managed_entry_store: lb_spec.managed_entry_store, driver: self)
+      lb_attributes = lb_options.delete(:load_balancer_attributes)
```
tyler-ball: Can you add a comment saying `We delete the options here because they are not valid in the create call, must be applied afterwords` - I could see myself asking why we are deleting this in about 2 months

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -269,6 +270,20 @@ def allocate_load_balancer(action_handler, lb_spec, lb_options, machine_specs)
         end
       end
 
+      # Update load balancer attributes
+      if lb_attributes
+        current = elb.client.describe_load_balancer_attributes(load_balancer_name: actual_elb.name)[:load_balancer_attributes]
+        desired = current.merge(lb_attributes)
```
tyler-ball: Ruby hash merging is weird - it doesn't do a recursive merge.  So it will depend on what AWS returns:

```ruby
current = {
  :access_log => {
    :enabled => false,
    :s3_bucket_name => "bucket"
  },
  :connection_settings => {
    :idle_timeout => 19
  }
}

lb_attributes = {
  :access_log => {
    :enabled => true
  }
}

# desired would be
desired = {
  :access_log => {
    :enabled => true
  },
  :connection_settings => {
    :idle_timeout => 19
  }
}
```

In this case we lose the bucket name.  I think we should do a recursive merge here.  Check out `deep_merge!` from https://github.com/chef/chef-provisioning-aws/blob/8558aa974d1ac7645c653ee5b6d711500b8dcc3b/lib/chef/provisioning/aws_driver/credentials.rb

dblessing: I didn't know that! Thanks.

dblessing:
@tyler-ball I made all the changes you suggested. Wow, merging/duplicating hashes in Ruby is a pain.

PR #201, opened 2015-05-15
replace instance.status with block that will wait until the instance …

jslusher:
…is ready to have a status


tyler-ball:
This was merged in another PR - Thanks @jslusher !

Issue #154, opened 2015-03-26
Certain security group rules definitions don't match correctly (not idempotent)


dblessing:
I chatted with @jkeiser on Gitter about this issue. Here are the specifics:

Given a SG:
```
aws_security_group "NATSG" do
  vpc "North_Virginia"
  inbound_rules [
    {:port => 22, :protocol => :tcp, :sources => ['10.0.128.0/20'] },
    {:port => 22, :protocol => :tcp, :sources => ['71.8.245.0/24'] },
    {:port => 22, :protocol => :tcp, :sources => ['173.190.236.0/24'] },
    {:port => 22, :protocol => :tcp, :sources => ['192.168.18.0/24'] },
    {:port => 80, :protocol => :tcp, :sources => ['10.228.0.0/16'] },
    {:port => 443, :protocol => :tcp, :sources => ['10.228.0.0/16'] },
    {:port => -1, :protocol => :icmp, :sources => ['0.0.0.0/0'] }
  ]
  outbound_rules [
    {:port => 80, :protocol => :tcp, :destinations => ['0.0.0.0/0'] },
    {:port => 443, :protocol => :tcp, :destinations => ['0.0.0.0/0'] },
    {:port => -1, :protocol => :icmp, :destinations => ['0.0.0.0/0'] }
  ]
  description "NAT Instance Security Group"
end
```

We get errors like:
```
================================================================================
    Error executing action `create` on resource 'aws_security_group[NATSG] (sg-xxxxxx)'
    ================================================================================

    AWS::EC2::Errors::InvalidPermission::Duplicate
    ----------------------------------------------
    the specified rule "peer: 0.0.0.0/0, TCP, from port: 80, to port: 80, ALLOW" already exists
```

This is because we specify the port as a single value (80) but it get translated to 80..80 on the back side. This messed with comparison.  @jkeiser said this about how it might be fixed:

"(We need to fix this too, obviously—if you want to do that, you can go to get_port_ranges(port_spec) and change the lines to do a range, line this. We should also add a check to ensure :port is there"
```
 when Hash
  if port_spec[:protocol]
    [ { port_range: port_spec[:port_range] || port_spec[:port]..port_spec[:port], protocol: port_spec[:protocol].to_s.to_sym } ]
  else
    get_port_ranges(port_spec[:port_range] || port_spec[:port]..port_spec[:port])
  end
```

nspottsie:
[...]

PR #237, opened 2015-06-22
Fix security group rule comparison

dblessing:
This fixes a very specific bug in `aws_security_group` rule comparison. If a protocol is specified, such as `protocol: -1` then `port_range` comparison would fail. This is because the desired rule would specify `port_range: -1..-1` while the actual rule would come back as `port_range: nil`. 

This is easily resolved by setting `port_range` to `-1..-1` in the absence of a specific `rule[:from_port]`. I also added a test that should catch this in the future. I wasn't able to explicitly test for the output of `port_range` since it's not available in the `ip_permission_list`. However, by just having the test in here we will get the following error if this happens again:

```
     AWS::EC2::Errors::InvalidPermission::Duplicate:
       aws_security_group[test_sg] (basic_chef_client::block line 96) had an error: AWS::EC2::Errors::InvalidPermission::Duplicate: the specified rule "peer: 0.0.0.0/0, ALL, ALLOW" already exists
```

Addresses the issue described in https://github.com/chef/chef-provisioning-aws/issues/154#issuecomment-112078637 (issue comment only - does not address the original intent of the issue report)

### DIFF
--- a/lib/chef/provider/aws_security_group.rb
+++ b/lib/chef/provider/aws_security_group.rb
@@ -148,7 +148,7 @@ class Chef::Provider::AwsSecurityGroup < Chef::Provisioning::AWSDriver::AWSProvi
     actual_rules = {}
     actual_rules_list.each do |rule|
       port_range = {
-        port_range: rule[:from_port] ? rule[:from_port]..rule[:to_port] : nil,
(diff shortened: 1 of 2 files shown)


tyler-ball:
Assigning myself as reviewer - I'll look at this soon.  Thanks @dblessing !

tyler-ball:
This looks good!

PR #198, opened 2015-05-08
Creating aws_network_acl resource, fixes #168

tyler-ball:
Fixes https://github.com/chef/chef-provisioning-aws/issues/168

:construction: This is till a WIP - needs update functionality, tests and README updated

Currently create and destroy are working

PR #241, opened 2015-06-27
Network acl

dblessing:
A continuation of #198 - started by @tyler-ball 

- [x] Rule create, update, delete
- [x] Subnet associations
- [x] Tests
- [x] Allow Array or Hash for `inbound/outbound_rules`?

One note - AWS SDK v1 API only supports linking ACLs from the subnet. I added the support on the `aws_subnet` object and added tests for that, too.

Example:
```
aws_network_acl 'ref-acl' do
  vpc 'ref-vpc'
  inbound_rules(
    { rule_number: 100, action: :allow, protocol: -1, cidr_block: '0.0.0.0/0' },
    { rule_number: 200, action: :allow, protocol: 6, port_range: 443..443, cidr_block: '172.31.0.0/24' }
  )
  outbound_rules(
    { rule_number: 100, action: :allow, protocol: -1, cidr_block: '0.0.0.0/0' }
  )
end
```

Changes to `aws_subnet`:
```
aws_subnet 'ref-subnet' do
  vpc 'ref-vpc'
  cidr_block '10.0.0.0/24'
  availability_zone az
  map_public_ip_on_launch true
  route_table 'ref-route-table'
  network_acl 'ref-acl'
end
```


lib/chef/provider/aws_network_acl.rb
```diff
@@ -0,0 +1,126 @@
...
+  def apply_rules(network_acl)
+    current_rules = network_acl.entries.map { |entry| entry_to_hash(entry) }
+    inbound_rules = new_resource.inbound_rules
+    outbound_rules = new_resource.outbound_rules
+    # AWS requires a deny all rule at the end. Delete here so we don't
+    # try to compare.
+    current_rules.delete_if { |rule| rule[:rule_number] == 32767 }
+
+    current_inbound_rules = current_rules.select { |rule| rule[:egress] == false }
+    desired_inbound_rules = inbound_rules ? inbound_rules.map { |rule| rule[:egress] = false; rule } : []
```
tyler-ball: One thing we should check for here is the difference between not specifying the `inbound_rules` attribute and specifying it as `inbound_rules []`.  This has come up recently on the Security Group resource.

In the first case (not specified) we don't change existing inbound rules at all.  In the second case (empty) we delete all inbound rules.

lib/chef/resource/aws_subnet.rb
```diff
@@ -75,6 +76,12 @@ class Chef::Resource::AwsSubnet < Chef::Provisioning::AWSDriver::AWSResourceWith
   #
   attribute :route_table#, kind_of: [ String, AwsRouteTable, AWS::EC2::RouteTable ], equal_to: [ :default_to_main ]
 
+  #
+  # The Network ACL to associate with this subnet. Subnets may only
+  # be associated with a single Network ACL.
+  #
+  attribute :network_acl, kind_of: [ String, AwsNetworkAcl, AWS::EC2::NetworkACL ]
```
tyler-ball: Another issue that has come up recently is the ability to remove references.  IE, how do I unlink a network_acl from a subnet?  I'm going to propose that we introduce the idea of adding the attribute value `network_acl false` as the signifier for unlinking the two.  How does this sound?

If we add this we should also add a test covering this.

Paging @jkeiser @randomcamel @metadave

tyler-ball:
Once again @dblessing this looks FANTASTIC!  Thank you for the awesome contribution.  I have two requests but I think they should be easy to accommodate.

dblessing:
Yeah, those suggestions shouldn't be a problem. I'll work this up in the next few days.

spec/integration/aws_network_acl_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+            vpc_id: test_vpc.aws_object.id,
+            entry_set:
+              [
+                { :rule_number=>500, :protocol=>"-1", :rule_action=>"allow", :egress=>true, :cidr_block=>"0.0.0.0/0" },
+                { :rule_number=>32767, :protocol=>"-1", :rule_action=>"deny", :egress=>true, :cidr_block=>"0.0.0.0/0" },
+                { :rule_number=>100, :protocol=>"-1", :rule_action=>"deny", :egress=>false, :cidr_block=>"10.0.0.0/24" },
+                { :rule_number=>32767, :protocol=>"-1", :rule_action=>"deny", :egress=>false, :cidr_block=>"0.0.0.0/0" }
+              ]
+          ).and be_idempotent
+        end
```
dblessing: I need some help on this test. Everything works (it's definitely checking to make sure the rules match) but the test fails because `.updated?` is false, not true. Is there an easy way to tell the test that's what we want?

lib/chef/resource/aws_subnet.rb
dblessing: On further investigation I believe the SDK is going to limit us here. There's no ability to simply disassociate a network acl. I even tried setting `network_acl = nil` instead of an id but it gave me an error. If anyone has ideas on how we might do that, I'm open to suggestions.

spec/integration/aws_network_acl_spec.rb
tyler-ball: This is failing because the `create_an_aws_object` matcher is hard-wired to listed for resource updates.  It wants to ensure the chef_run which occurred created a resource.

I think the right fix here is to create a `matches_an_aws_object` matcher that doesn't listen to resource updates.  Instead, it will just validate that the resource you specify (`'test_network_acl'`) exists and has the attributes that you specify.  I'll get a PR out shortly.

lib/chef/resource/aws_subnet.rb
tyler-ball: From Office Hours - @dblessing suggested we could try looking up the default network ACL that subnets belong to and exchange that link with the current one when a `nil` attribute is specified.  

If it is SUPER easy to do this, go ahead.  But if its not just add a TODO explaining the issue and we can tackle it at another time.  If your use case doesn't call for it, we can wait until someone needs it.

dblessing:
I opted for the 'TODO' on the network_acl disassociation. I don't think I have the time right now. Thanks again for your help with the tests @tyler-ball - it worked great. All integration tests pass for NACLs and Subnets now.

PR #248, opened 2015-07-02
Adding a matcher which doesn't do any CRUD operations

tyler-ball:
It simply tries to match aws object with expected values.

\cc @jkeiser @randomcamel @dblessing 

To support https://github.com/chef/chef-provisioning-aws/pull/241


dblessing:
:+1:

PR #263, opened 2015-07-27
Adding exponential backoff when checking taggable status

tyler-ball:
Some of the tests were failing and this seems to help.  But it is also a good idea.

Do you think we should cap the timeout at 8 or 16 seconds and stop doubling it there?

\cc @fnichol @jkeiser @randomcamel

### DIFF
--- a/lib/chef/provider/aws_vpc.rb
+++ b/lib/chef/provider/aws_vpc.rb
@@ -5,6 +5,8 @@ require 'retryable'
 
 class Chef::Provider::AwsVpc < Chef::Provisioning::AWSDriver::AWSProvider
 
+  class NeverObtainedExistence < RuntimeError; end
+
   def action_create
     vpc = super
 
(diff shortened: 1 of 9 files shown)


lib/chef/provider/aws_vpc.rb
```diff
@@ -165,7 +165,7 @@ def update_internet_gateway(vpc)
       if !current_ig
         converge_by "attach new Internet Gateway to VPC #{vpc.id}" do
           current_ig = AWS.ec2(config: vpc.config).internet_gateways.create
-          Retryable.retryable(:tries => 15, :sleep => 1, :matching => /never obtained existence/) do
+          Retryable.retryable(:tries => 15, :sleep => lambda { |n| 2**n }, :matching => /never obtained existence/) do
```
randomcamel: Why not replace this `retry_with_backoff` as above?

randomcamel:
I'd be fine with a final timeout of 64 seconds in general--maybe more if the user were to specify it. I'm not sure there's an empirical way to arrive at a number.

lib/chef/provisioning/aws_driver/aws_provider.rb
```diff
@@ -276,7 +276,7 @@ def wait_for_status(aws_object, expected_status, acceptable_errors = [], tries=6
     expected_status = [expected_status].flatten
     current_status = aws_object.status
 
-    Retryable.retryable(:tries => tries, :sleep => sleep, :on => StatusTimeoutError) do |retries, exception|
```
tyler-ball: This change doesn't include a backoff.  It only changes it so we catch 'id not found' errors and retry those too

PR #269, opened 2015-07-30
Replacing use_private_ip_for_ssh with transport_address_location, fixes #267

### LINKED ISSUE #267: Can't connect to EC2 instance in VPC with public IP
Copied from https://github.com/chef/chef-provisioning/issues/385
\cc @dvlscm

------

I spin up a box which has a public IP and a public DNS. It's in the public subnet of the default VPC. I can SSH into the machine manually from my command line (Mac OS X) using the public DNS, but I can _not_ connect with the public IP. So that's the baseline scenario.

Looks to me like chef-provisioning creates an the EC2 instance I want, then it tries to SSH into the instance using `ubuntu@[public ip address]`, which won't fly.

Is there a way to tell chef-provisioning to use the public DNS instead when it tries to SSH into machines to converge?

tyler-ball:
Fixes #267 

It should be clear from the README and CHANGELOG what the update is.  If it is not, let me know!

\cc @fnichol @randomcamel @jkeiser


afiune:
I totally say :shipit: I like it!

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -1019,7 +1033,13 @@ def create_servers(action_handler, specs_and_options, parallelizer, &block)
             instance.source_dest_check = machine_options[:source_dest_check] if machine_options.has_key?(:source_dest_check)
             converge_tags(instance, machine_options[:aws_tags], action_handler)
             machine_spec.reference['key_name'] = bootstrap_options[:key_name] if bootstrap_options[:key_name]
-            %w(is_windows ssh_username sudo use_private_ip_for_ssh ssh_gateway).each do |key|
+            # TODO 2.0 We no longer support `use_private_ip_for_ssh`, only `transport_address_location`
+            if machine_options[:use_private_ip_for_ssh]
+              Chef::Log.warn("The machine_option ':use_private_ip_for_ssh' has been deprecated, use ':transport_address_location'")
```
jkeiser: Two things:

1. We should really only warn this once per run
2. The suggestion should probably be "use 'transport_address_location: :private_ip'"

tyler-ball: Do you have a good pattern example of a way to only warn once per run?  Just a class variable?

jkeiser: In this case once per driver is probably fine.

PR #270, opened 2015-07-31
only update bootstrap_options[:user_data] in windows, if one hasn't been provided

brumschlag:
bootstrap_options[:user_data] always gets overwritten even if the user has provided their own.


tyler-ball:
Right you are!  Thanks for this PR

PR #272, opened 2015-08-03
Initial commit of aws_rds_instance resource

stevendanna:
This is a bare-bones resource for creating and destroying RDS instances.
RDS instances have a lot of possible options, so rather than list them
all in the resources, I've provided an `additional_options` hash that
the user can use to directly pass options to the underlying driver.  We
will likely expand the functionality here as we use it ourselves.


tylercloke:
:+1: Looks pretty straightforward, let's just make sure we document this with the other resources.

lib/chef/resource/aws_rds_instance.rb
```diff
@@ -0,0 +1,33 @@
+require 'chef/provisioning/aws_driver/aws_resource'
+
+
+class Chef::Resource::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSResource
+  aws_sdk_type AWS::RDS::DBInstance
+
+  attribute :db_instance_identifier, kind_of: String, name_attribute: true
```
tyler-ball: Awesome - this is unique across the account and region.  That means we don't have to create this class as a `Chef::Provisioning::AWSDriver::AWSResourceWithEntry`

lib/chef/resource/aws_rds_instance.rb
```diff
@@ -0,0 +1,33 @@
+require 'chef/provisioning/aws_driver/aws_resource'
+
+
+class Chef::Resource::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSResource
+  aws_sdk_type AWS::RDS::DBInstance
```
tyler-ball: `aws_sdk_type AWS::RDS::DBInstance, id: db_instance_identifier`

This means that if you create a resource like:

```ruby
instance = rds.db_instances['foo']
aws_rds_instance instance do
  iops 2000
end
```

It should correctly create the resource when the `name` attribute is a `AWS::RDS::DBInstance` object instead of a String.

stevendanna: ah, great!

lib/chef/resource/aws_rds_instance.rb
tyler-ball: This should also have `aws_id_attribute: true` appended to the list of options

lib/chef/resource/aws_rds_instance.rb
```diff
@@ -0,0 +1,33 @@
...
+  attribute :db_instance_class, kind_of: String
+  attribute :multi_az, default: false, kind_of: [TrueClass, FalseClass]
+  attribute :allocated_storage, kind_of: Integer, default: 5
+  attribute :iops, kind_of: Integer
+  attribute :publicly_accessible, kind_of: [TrueClass, FalseClass], default: false
+  attribute :master_username, kind_of: String
+  attribute :master_user_password, kind_of: String
+  attribute :db_name, kind_of: String
+  attribute :db_port, kind_of: Integer
+
```
tyler-ball: So, our tagging support now is kind of crappy.  Most AWS clients have different methods of tagging the resources that client is responsible for and I haven't refactored the way we do it very well.  But the first step is to add

```ruby
  # @param aws_tags [Hash] Should be a hash of keys & values to add.  Keys and values
  #        can be provided as symbols or strings, but will be stored in AWS as strings.
  attribute :aws_tags, kind_of: Hash
```

Unfortunately right now this is copied from `AWSResourceWithEntry`.  I need to clean this up and create a module that represents taggability.

stevendanna: The RDS API supports tagging on creation, so we can thread this through to the creation call pretty easily.

tyler-ball: The [creates](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/RDS/Client.html#create_db_instance-instance_method) are fairly easy, but its the updates that I am more worried about.

The [`modify_db_instance`](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/RDS/Client.html#modify_db_instance-instance_method) method does not accept tags - they must be managed after creation with [`add_tags_to_resource`](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/RDS/Client.html#add_tags_to_resource-instance_method) and [`remove_tags_from_resource`](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/RDS/Client.html#remove_tags_from_resource-instance_method).

We have logic in [`aws_provider`](https://github.com/chef/chef-provisioning-aws/blob/3635d8853f6169702c679cbd157a9dc46e2c8860/lib/chef/provisioning/aws_driver/aws_provider.rb#L237-L264) for updating the tags but it only works for certain aws objects.

My idea is to create a module which contains that same basic logic (look at current, diff with desired, add/update/delete as necessary) but has a bunch of setters with defaults so the idea of 'how do I read existing tags' becomes configurable.  Then resource authors would only need to include this module and configure it correctly to gain the benefit of tags.

lib/chef/provider/aws_rds_instance.rb
```diff
@@ -0,0 +1,35 @@
+require 'chef/provisioning/aws_driver/aws_provider'
+
+class Chef::Provider::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSProvider
+  REQUIRED_OPTIONS = %w(db_instance_identifier allocated_storage engine
+                        db_instance_class master_username master_user_password)
+
+  OTHER_OPTIONS = %w(engine_version multi_az iops publicly_accessible db_name db_port)
+
+  def update_aws_object(instance)
+    Chef::Log.warn("aws_rds_instance does not support modifying a started instance")
```
tyler-ball: Is this true?  I don't see anything about it on http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/RDS/Client.html#modify_db_instance-instance_method

stevendanna: I meant this to mean that this resource doesn't support it yet.

lib/chef/provider/aws_rds_instance.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def create_options
+    opts = {}
+    REQUIRED_OPTIONS.each do |opt|
+      opts[opt] = new_resource.send(opt.to_sym)
+    end
+    OTHER_OPTIONS.each do |opt|
+      opts[opt] = new_resource.send(opt.to_sym) if ! new_resource.send(opt.to_sym).nil?
+    end
+    opts.merge(new_resource.additional_options)
```
tyler-ball: I would think we would want the discrete attributes (like `iops`) to overwrite the hash in `additional_options`.  With the current code, if I specify both `iops 100` and `additional_options {iops: 200}` then the `200` value will win.  I would expect the `100` value to win.

Either way we decide, we should document this behavior.

lib/chef/provider/aws_rds_instance.rb
tyler-ball: I'm fine with that - if you won't need updates we can merge this PR without support for it.  We can add it as soon as we need it.  Which we should do preferably before we have stuff in production being managed :)

stevendanna: Cool, I plan on circling back around to adding update support to this once @tylercloke and I are able to stand up all the resources we need.

lib/chef/resource/aws_rds_instance.rb
stevendanna: Tagging added on create, since we don't yet support modify.

lib/chef/provider/aws_rds_instance.rb
tylercloke: We should be fine without it for now.

lib/chef/provider/aws_rds_instance.rb
```diff
@@ -0,0 +1,49 @@
...
+  def create_options
+    opts = {}
+    opts.merge(new_resource.additional_options)
+    REQUIRED_OPTIONS.each do |opt|
+      opts[opt] = new_resource.send(opt.to_sym)
+    end
+    OTHER_OPTIONS.each do |opt|
+      opts[opt] = new_resource.send(opt.to_sym) if ! new_resource.send(opt.to_sym).nil?
+    end
+    opts[:tags] = hash_to_tag_array(new_resource.aws_tags) if new_resource.aws_tags
```
tyler-ball: You shouldn't _need_ to do this because the provider [takes care](https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/aws_provider.rb#L124) of it for you.

Pros of this - having this here reduces the number of API calls we make to AWS
Cons of this - It adds more duplicated tagging code, of which there is already a lot ([1](https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/aws_provider.rb#L237), [2](https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/driver.rb#L94-L101), [3](https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/driver.rb#L284-L311), [4](https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/driver.rb#L1080))

Because my primary concern at this point is code complexity I would rather remove this logic and let the provider handle tagging.  Unless that wasn't working somehow?

lib/chef/provider/aws_rds_instance.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def destroy_aws_object(instance)
+    converge_by "Deleting RDS instance #{new_resource.db_instance_identifier} in #{region}" do
+      instance.delete(skip_final_snapshot: true)
+    end
+  end
+
+  def create_options
+    opts = {}
+    opts.merge(new_resource.additional_options)
```
tyler-ball: Same note from https://github.com/chef/chef-provisioning-aws/pull/272#discussion_r36194928

tyler-ball:
Ditto from https://github.com/chef/chef-provisioning-aws/pull/273#issuecomment-128450690 and the README needs a resource section added for this explaining how the `additional_options` gets merged.

lib/chef/provider/aws_rds_instance.rb
tylercloke: As tested for, the explicit value overrides `additional_options` (see [test](https://github.com/chef/chef-provisioning-aws/blob/08e5c2d3eed5f93550973167a6fce554e3a8ab38/spec/integration/aws_rds_instance_spec.rb#L69-L74)).

I'll add some documentation.

lib/chef/resource/aws_rds_instance.rb
```diff
@@ -0,0 +1,36 @@
...
+class Chef::Resource::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSResource
+  aws_sdk_type AWS::RDS::DBInstance, id: :db_instance_identifier
+
+  attribute :db_instance_identifier, kind_of: String, name_attribute: true
+
+  attribute :engine, kind_of: String
+  attribute :engine_version, kind_of: String
+  attribute :db_instance_class, kind_of: String
+  attribute :multi_az, default: false, kind_of: [TrueClass, FalseClass]
+  attribute :allocated_storage, kind_of: Integer, default: 5
```
tyler-ball: A common pattern we see with chef-provisioning is someone declaring a resource with all the possible attributes (like `allocated_storage 5000`) and then updating just one resource on the attribute later (like `description "new description"`).

If we default the storage to 5 then when they have their second resource invocation later, it will reset the storage from 5000 down to 5.  This is also a consequence of the fact that chef-provisioning-aws resources don't perform resource cloning.

I'll remove this in the commit I am working on

lib/chef/provider/aws_rds_instance.rb
stevendanna: I don't think coverge_tags will work.  It does the following:

https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/aws_provider.rb#L256

But tags isn't even a method on AWS::RDS::DBInstance:

```
irb(main):001:0> require 'aws'
=> true
irb(main):002:0> r = AWS::RDS::DBInstance.new("foo", {})
=> <AWS::RDS::DBInstance db_instance_identifier:foo>
irb(main):003:0> r.tags
NoMethodError: undefined method `tags' for <AWS::RDS::DBInstance db_instance_identifier:foo>:AWS::RDS::DBInstance
        from (irb):3
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/lib/bundler/cli/console.rb:38:in `run'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/lib/bundler/cli.rb:286:in `console'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/lib/bundler/vendor/thor/command.rb:27:in `run'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/lib/bundler/vendor/thor/invocation.rb:121:in `invoke_command'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/lib/bundler/vendor/thor.rb:363:in `dispatch'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/lib/bundler/vendor/thor/base.rb:440:in `start'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/lib/bundler/cli.rb:9:in `start'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/bin/bundle:20:in `block in <top (required)>'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/lib/bundler/friendly_errors.rb:5:in `with_friendly_errors'
        from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/bundler-1.7.12/bin/bundle:18:in `<top (required)>'
        from /opt/chefdk/embedded/bin/bundle:23:in `load'
        from /opt/chefdk/embedded/bin/bundle:23:in `<main>'
```

lib/chef/resource/aws_rds_instance.rb
stevendanna: @tyler-ball Heya, was looking at these comments.  Gonna hold off on more changes until you commit.  I think @tylercloke has addressed the documentation needs.

stevendanna:
:+1: :shipit:

Fodoj:
@tyler-ball when do you plan to release 1.4.0? I really need rds support already and I can't get it from 1.3.1 :( Can I help in speeding up this release?

Fodoj:
@tyler-ball or hire me, so I can work on it full-time and add support for every aws resource in next few months :)

tyler-ball:
Its definitely on my roadmap to do soon.  Do you use bundler or the ChefDK to manage your provisioning workstation?  There are workarounds to get the latest code

Fodoj:
@tyler-ball I have a separate provisioner node on ec2 ¯\_(ツ)_/¯

tyler-ball:
Your EC2 node is what I call the 'provisioner workstation'.  Do you do `gem install chef-provisioning` or do you install the ChefDK or something else?  How do you get the provisioning code on your EC2 node?

Fodoj:
@tyler-ball via my chef cookbook called "provisioner", of course :) Among other things, it has this:

```ruby
chef_gem "chef-provisioning" do
  version "1.3.0"
end

chef_gem "cheffish" do
  version "1.3.1"
end

chef_gem "chef-provisioning-aws"
```

Fodoj:
I use chef provisioner to provision a provisioner node that provisions other nodes, yes.

tyler-ball:
Yo Dawg!  Well, I'm scratching my head trying to figure out how to get Chef to install a gem to `chef_gem` from a github source.  You could try `source: https://github.com/chef/chef-provisioning-aws` but I don't think that will work.

Otherwise you could use the [git resource](http://docs.chef.io/resource_git.html) and a `ruby_block` to execute `gem build` on that directory, then point the `source` attribute to that built gem.  If you do this you'll probably have to uninstall previously installed gems or force an install because the version in master doesn't change often.

tyler-ball:
But yes, I agree, we need to get these released soon :)

PR #273, opened 2015-08-03
Initial commit of aws_cloudsearch_domain resource

stevendanna:


lib/chef/resource/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,46 @@
+require 'chef/provisioning/aws_driver/aws_resource'
+
+module AWS
+  class CloudSearch
+    class Domain
+      # The version of the AWS sdk we are using doesn't have a model
```
tyler-ball: The `aws_cache_*` resources had this same problem.  For those we specified `aws_sdk_type AWS::ElastiCache`  We could do the same thing here unless we find an issue with it - `aws_sdk_type AWS::CloudSearch`

lib/chef/resource/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,46 @@
...
+      # object for CloudSearch Domains. This empty class is here to
+      # make the reset of chef-provisioning happy.
+    end
+  end
+end
+
+class Chef::Resource::AwsCloudsearchDomain < Chef::Provisioning::AWSDriver::AWSResource
+  aws_sdk_type ::AWS::CloudSearch::Domain
+  attribute :domain_name, kind_of: String, name_attribute: true
```
tyler-ball: `aws_id_attribute: true`

lib/chef/resource/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,46 @@
...
+  # TODO(ssd): Like Access Polcies, we should decide
+  # whether we want a DSL for defining index fields, or just allow the
+  # user to pass in an array properly formated hash.
+  attribute :index_fields, kind_of: Array
+
+  # None of the cloudsearch objects actually have instance-specific
+  # objects in the version of the AWS API we are using.  This will
+  # return a hash with some relevant information about the domain.
+  def aws_object
+    driver.cloudsearch.describe_domains[:domain_status_list].find {|d| d[:domain_name] == name && !d[:deleted]}
```
tyler-ball: Looking at http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/CloudSearch/Client.html#describe_domains-instance_method I think you could do

```ruby
driver.cloudsearch.describe_domains(:domain_names => domain_name)[:domain_status_list].find {|d| !d[:deleted]}
```

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -534,6 +534,12 @@ def destroy_machine(action_handler, machine_spec, machine_options)
       strategy.cleanup_convergence(action_handler, machine_spec)
     end
 
+    def cloudsearch(api_version="20130101")
+      @rds ||= {}
```
tyler-ball: `@cloudsearch` instead of `@rds`

lib/chef/resource/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,46 @@
...
+      # object for CloudSearch Domains. This empty class is here to
+      # make the reset of chef-provisioning happy.
+    end
+  end
+end
+
+class Chef::Resource::AwsCloudsearchDomain < Chef::Provisioning::AWSDriver::AWSResource
+  aws_sdk_type ::AWS::CloudSearch::Domain
+  attribute :domain_name, kind_of: String, name_attribute: true
+  attribute :cloudsearch_api_version, equal_to: ["20130101", "20110201"], default: "20130101"
```
tyler-ball: I could see a future version breaking compatibility.  But for now would we ever use anything besides `20130101`?

tyler-ball: Ahh - I see why you did this.  Their [docs](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/CloudSearch/Client.html) specify that 2011 is the default because 2013 is not backwards compatiable.

stevendanna: If the domain was created against 20110201 I believe you have to continue to access it with this API.  Unfortunately, I can't test this because it seems like our AWS account isn't even allowed to make requests via that API version.

tyler-ball: I still think we should remove this as a resource attribute and hard code it in the driver.  Why would we allow cookbook authors to use the old API?

stevendanna: I'm fine with that, it just means you can't use this to manage already created search domains.

tyler-ball: Ahh - I didn't know that you couldn't manage 2011 objects with the 2013 API.  Okay, ignore my original statement - lets leave it in with the smart default it currently has.  We should also document this in the README.

lib/chef/provisioning/aws_driver/driver.rb
stevendanna: Ha! Thanks

lib/chef/provider/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,126 @@
...
+
+class Chef::Provider::AwsCloudsearchDomain < Chef::Provisioning::AWSDriver::AWSProvider
+
+  def create_aws_object
+    converge_by "Creating new CloudSearch domain #{new_resource.domain_name}" do
+      create_domain
+    end
+    # multi_az defaults to false on creation
+    # so we only need to hit the API if it is true
+    if new_resource.multi_az
```
tyler-ball: Lets DRY this up by just calling `update_aws_object(instance)`.  The `instance` should be returned from the `create_domain` method.

lib/chef/provider/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,126 @@
...
+      end
+    end
+
+    if update_index_fields?
+      Chef::Log.warn("Updating existing index_fields not currently supported")
+    end
+  end
+
+
+  # TODO(ssd): Expand these to be idempotent?  It takes an extra API
```
tyler-ball: I vote to make them idempotent.  After the create is finished we will know the current state.  And the `update_aws_object` is passed the current aws object, so this query will have already been made.  These `?` methods just need to take in that existing object for comparison with `new_resource`.

Man, this will be sooooooo much easier with resources 2.0

lib/chef/provider/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,126 @@
...
+  def update_policy?
+    ! new_resource.access_policies.nil?
+  end
+
+  def update_index_fields?
+    ! new_resource.index_fields.nil?
+  end
+
+  def create_domain
+    cs_client.create_domain(domain_name: new_resource.domain_name)
```
tyler-ball: Per http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/CloudSearch/Client.html#create_domain-instance_method I think this will need to inspect the `:domain_status` field to get the returned object.  That should always be `in_process` or whatever the actual status is.

lib/chef/provider/aws_cloudsearch_domain.rb
stevendanna: I started in on making this idempotent and remembered what I meant by "extra API call".  The describe_domain api call that we use to get the object doesn't return availability, scalability, indexing, or access policy info.  Each piece of that info is a separate call:

```
irb(main):023:0> pp AWS::CloudSearch::Client::V20130101.new.describe_domains(domain_names: ["foo"])[:domain_status_list].first; nil
{:created=>true,
 :requires_index_documents=>false,
 :domain_id=>"862552916454/foo",
 :doc_service=>nil,
 :search_partition_count=>0,
 :deleted=>false,
 :search_service=>nil,
 :domain_name=>"foo",
 :processing=>true,
 :search_instance_count=>0,
 :arn=>"arn:aws:cloudsearch:us-east-1:862552916454:domain/foo",
 :limits=>
  {:available_instance_types=>{:member=>"search.m3.2xlarge"},
   :maximum_replication_count=>5,
   :maximum_partition_count=>10}}
irb(main):024:0> pp AWS::CloudSearch::Client::V20130101.new.describe_availability_options(domain_name: "foo")
{:availability_options=>{:options=>false, :status=>{:update_date=>2015-08-04 23:27:05 UTC, :state=>"Processing", :creation_date=>2015-08-04 23:27:05 UTC, :update_version=>2, :pending_deletion=>false}}, :response_metadata=>{:request_id=>"9b8826fc-3b03-11e5-90cc-a1231d3e2650"}}
```

lib/chef/resource/aws_cloudsearch_domain.rb
stevendanna: I am still digging through what this actually does, but adding it breaks this module under test.  If I add it to to the relevant attribute in aws_cache_subnet_group, it also breaks those tests, so my guess is that this only works for things with an actual domain object.

lib/chef/resource/aws_cloudsearch_domain.rb
stevendanna: I think it is better to keep this for now.  My problem with using an existing class is that future code might make assumptions about this being an actual AWS model object, and if it does I want it to blow up loudly.  If we use another object in the system we don't control, it might silently "work" for some time and produce hard-to-debug bugs.

stevendanna:
@tyler-ball I have addressed most of your comments and left comments for the two I didn't address.

I went down the road of making more complete integration tests, but unless cloudsearch gets a lot faster, I'm not sure it is realistic.  In my testing it can take a VERY long time for many of the options that you set to actually show up in the API calls that are supposed to return their value.  I think we could get around this by writing a complete model object for this CloudSearchDomain's that is able to understand all of the various states of "partial update" that the domain might be in.  Unfortunately, I think that is a bit out of scope for us right now since we haven't started using cloudsearch yet and thus don't necessarily understand what users of this will expect.  I'd rather not pin us into an API until we have some experience with this resource.

lib/chef/resource/aws_cloudsearch_domain.rb
tyler-ball: I'm fine with this.  It seems like a better pattern, so thanks!

lib/chef/resource/aws_cloudsearch_domain.rb
tyler-ball: It enables logic where if you have

```ruby
aws_cloudsearch_domain aws_object do
  ...
end
```

Where `aws_object` is an AWS model that you have gotten from somewhere, then the resource will still converge successfully.  But because cloud search doesn't have an AWS model then I think its okay to leave it off for now.

lib/chef/provider/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,160 @@
...
+  def create_aws_object
+    domain = nil # define here to ensure it is available outside of the coverge_by scope
+    converge_by "Creating new CloudSearch domain #{new_resource.domain_name}" do
+      domain = create_domain
+    end
+
+    update_aws_object(domain)
+  end
+
+  def destroy_aws_object(instance)
```
tyler-ball: The word `instance` actually refers to an AWS [instance](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/EC2/Instance.html).  I think you should replace `instance` with `domain`

lib/chef/provider/aws_cloudsearch_domain.rb
```diff
@@ -0,0 +1,160 @@
...
+    converge_by "Deleting CloudSearch domain #{new_resource.domain_name}" do
+      cs_client.delete_domain(domain_name: new_resource.domain_name)
+    end
+  end
+
+  def update_aws_object(instance)
```
tyler-ball: Ditto on https://github.com/chef/chef-provisioning-aws/pull/273/files#r36438166

spec/integration/aws_cloudsearch_domain_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'spec_helper'
+
+describe Chef::Resource::AwsCloudsearchDomain do
+  extend AWSSupport
+  when_the_chef_12_server "exists", organization: "foo", server_scope: :context do
+    with_aws "when connected to AWS" do
+      it "aws_cloudsearch_domain 'test-cloudsearch-domain' creates a cloudsearch domain" do
```
tyler-ball: RE, your speed concerns:

You should tag this test as `:super_slow` like https://github.com/chef/chef-provisioning-aws/blob/master/spec/integration/machine_spec.rb#L16

tyler-ball:
Some minor comments - once those are addressed I am :+1: and will merge this

tyler-ball:
Actually, before we merge this it needs to be added to [ref_full.rb](https://github.com/chef/chef-provisioning-aws/blob/master/docs/examples/ref_full.rb) and [ref_destroy.rb](https://github.com/chef/chef-provisioning-aws/blob/master/docs/examples/ref_destroy.rb).  These are good examples for users of what the resource looks like when _all_ its attributes are populated.

lib/chef/resource/aws_cloudsearch_domain.rb
tyler-ball: On further though, lets also change this to just `name` - we already know we're within the Cloudsearch Domain

spec/integration/aws_cloudsearch_domain_spec.rb
adamedx: Despite slowness, good to see we're including tests.

stevendanna: This test isn't too slow.  The comment about slowness was with adding additional assertions since some data doesn't show up in the API for a *very* long time (on the order of 10s of minutes).

tyler-ball: Ahhhh, I understand - carry on then!

PR #276, opened 2015-08-04
Initial commit of aws_rds_subnet_group resource

stevendanna:
TODO:
- [x] `subnets` section of `create_an_aws_db_subnet_group` is matching on an ordered array, but what is returned by the resource is not ordered. How can we make it match unordered array (can't figure out how to use regular ol' rspec in there).
- [x] Obvious the AZs I have defined in the tests won't work across regions, but finding ones that do is non-trivial (aka "a" and "b" are not always AZs in every region).


lib/chef/resource/aws_db_subnet_group.rb
```diff
@@ -0,0 +1,20 @@
+require 'chef/provisioning/aws_driver/aws_resource'
+require 'chef/resource/aws_subnet'
+
+class Chef::Resource::AwsDbSubnetGroup < Chef::Provisioning::AWSDriver::AWSResource
+  aws_sdk_type AWS::RDS
```
tylercloke: Shouldn't this be returning the group and not the base RDS object?

stevendanna: Unfortunately there isn't a group object in the v1 SDK.  This is just here to make provisioning happy.  I should comment it.

lib/chef/provider/aws_db_subnet_group.rb
```diff
@@ -0,0 +1,86 @@
...
+      ret << "  set group description to #{desired_options[:db_subnet_group_description]}"
+    end
+
+    if ! xor_array(desired_options[:subnet_ids], subnet_ids(object[:subnets])).empty?
+      pp desired_options[:subnet_ids]
+      pp object[:subnets]
+      ret << "  set subnets to #{desired_options[:subnet_ids]}"
+    end
+
+    if ! (desired_options[:tags].nil? || desired_options[:tags].empty?)
```
stevendanna: This should be desired_options[:aws_tags] I believe.

tyler-ball:
I'm a bigger fan of `aws_rds_*` for all the RDS resources - was there a reason you chose `aws_db`?

stevendanna:
> I'm a bigger fan of aws_rds_* for all the RDS resources - was there a reason you chose aws_db?

@tyler-ball Mostly to stay in line with what I was seeing in the written documentation when setting it up manually.  Happy to change it to aws_rds_subnet_group.  I think that will be just as clear to anyone reading the documentation and help clarify what it is for.

lib/chef/provider/aws_db_subnet_group.rb
```diff
@@ -0,0 +1,86 @@
...
+  # If no updates are needed, an empty array is returned.
+  #
+  def required_updates(object)
+    ret = []
+    if desired_options[:db_subnet_group_description] != object[:db_subnet_group_description]
+      ret << "  set group description to #{desired_options[:db_subnet_group_description]}"
+    end
+
+    if ! xor_array(desired_options[:subnet_ids], subnet_ids(object[:subnets])).empty?
+      pp desired_options[:subnet_ids]
```
stevendanna: Yikes, we probably don't want this debugging code.

stevendanna:
Renamed as requested.

stevendanna:
@tylercloke Thanks for digging into these tests!  I think I fixed the ordering issue here:

https://github.com/chef/chef-provisioning-aws/commit/7711d94f778e52d7a5e75aad5bc98f06d2ef3a41

tyler-ball:
Can I call @jkeiser in for some support on the matchers?

tylercloke:
@tyler-ball ^ Squashed, rebased, and added ref_full.rb and ref_destroy.rb examples

docs/examples/aws_rds_subnet_group.rb
```diff
@@ -0,0 +1,21 @@
...
+
+subnet2 = aws_subnet "subnet_2" do
+  vpc 'coolvpc'
+  cidr_block "10.0.0.64/26"
+  availability_zone "us-east-1c"
+end
+
+aws_rds_subnet_group "db-subnet-group" do
+  db_subnet_group_description "some_description"
+  subnet_ids lazy {[subnet1.aws_object.id, subnet2.aws_object.id]}
```
tyler-ball: Does this not work if you specify `subnet_ids 'subnet', 'subnet_2'`?  It should...

tylercloke: I don't think you need the lazy block, but I don't think just using the strings worked. I can try in the tests again.

lib/chef/provider/aws_rds_subnet_group.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+    end
+  end
+
+  def desired_options
+    opts = {}
+    opts[:db_subnet_group_name] = new_resource.db_subnet_group_name
+    opts[:db_subnet_group_description] = new_resource.db_subnet_group_description
+    opts[:subnet_ids] = new_resource.subnet_ids
+    opts[:tags] = tag_hash_to_array(new_resource.aws_tags) if new_resource.aws_tags
```
tyler-ball: Ditto from https://github.com/chef/chef-provisioning-aws/pull/272/files#r36440811

lib/chef/provider/aws_rds_subnet_group.rb
```diff
@@ -0,0 +1,84 @@
...
+    ret = []
+    if desired_options[:db_subnet_group_description] != object[:db_subnet_group_description]
+      ret << "  set group description to #{desired_options[:db_subnet_group_description]}"
+    end
+
+    if ! xor_array(desired_options[:subnet_ids], subnet_ids(object[:subnets])).empty?
+      ret << "  set subnets to #{desired_options[:subnet_ids]}"
+    end
+
+    if ! (desired_options[:aws_tags].nil? || desired_options[:aws_tags].empty?)
```
tyler-ball: I don't want to block this PR on being able to update tags.  I think we could merge it without update support now and file an issue to improve it later.

But for that to work we'll have to make sure `aws_tags` gets unset on an update, otherwise https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/provisioning/aws_driver/aws_provider.rb#L245 will raise an exception

lib/chef/provider/aws_rds_subnet_group.rb
```diff
@@ -0,0 +1,84 @@
...
+  def destroy_aws_object(object)
+    converge_by "delete RDS Subnet Group #{new_resource.name} in #{region}" do
+      driver.delete_db_subnet_group(db_subnet_group_name: new_resource.db_subnet_group_name)
+    end
+  end
+
+  def update_aws_object(object)
+    updates = required_updates(object)
+    if ! updates.empty?
+      converge_by updates do
```
tyler-ball: I like making an update string so we get proper output, but don't we need a `updates.join("\n")` so the output appears on a new line?  Or does the `converge_by` handle this already?  What does the output look by when multiple attributes are updated?

lib/chef/provider/aws_rds_subnet_group.rb
```diff
@@ -0,0 +1,84 @@
+require 'chef/provisioning/aws_driver/aws_provider'
+
+class Chef::Provider::AwsRdsSubnetGroup < Chef::Provisioning::AWSDriver::AWSProvider
+
+  def create_aws_object
+    converge_by "create new RDS Subnet Group #{new_resource.name} in #{region}"do
```
tyler-ball: nit - space before the `do`

lib/chef/resource/aws_rds_subnet_group.rb
```diff
@@ -0,0 +1,20 @@
+require 'chef/provisioning/aws_driver/aws_resource'
+require 'chef/resource/aws_subnet'
+
+class Chef::Resource::AwsRdsSubnetGroup < Chef::Provisioning::AWSDriver::AWSResource
+  aws_sdk_type AWS::RDS
+
+  attribute :db_subnet_group_name, kind_of: String, name_attribute: true
+  attribute :db_subnet_group_description, kind_of: String
```
tyler-ball: These attributes should just be `name` and `description` - we already know we are within the RDS Subnet Group object

tylercloke: These are mapping directly to the required args that are used to create the subnet group:

http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/RDS/Client.html#create_db_subnet_group-instance_method

lib/chef/provider/aws_rds_subnet_group.rb
tyler-ball: Nope - output looks great!

tyler-ball:
Thanks for the awesome work @stevendanna and @tylercloke !

PR #283, opened 2015-08-09
edit strings for consistency, branding (AWS)

jamescott:
Consistent converge_by statements
Consistent capitalization
Applied AWS branding patterns to all terms

### DIFF
--- a/lib/chef/provider/aws_auto_scaling_group.rb
+++ b/lib/chef/provider/aws_auto_scaling_group.rb
@@ -6,7 +6,7 @@ class Chef::Provider::AwsAutoScalingGroup < Chef::Provisioning::AWSDriver::AWSPr
   protected
 
   def create_aws_object
-    converge_by "create new Auto Scaling Group #{new_resource.name} in #{region}" do
+    converge_by "create Auto Scaling group #{new_resource.name} in #{region}" do
       options = desired_options.dup
       options[:min_size] ||= 1
       options[:max_size] ||= 1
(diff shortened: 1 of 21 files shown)


lib/chef/provider/aws_auto_scaling_group.rb
```diff
@@ -6,7 +6,7 @@ class Chef::Provider::AwsAutoScalingGroup < Chef::Provisioning::AWSDriver::AWSPr
   protected
 
   def create_aws_object
-    converge_by "create new Auto Scaling Group #{new_resource.name} in #{region}" do
+    converge_by "create Auto Scaling group #{new_resource.name} in #{region}" do
```
dblessing: I love the consistency of changing all the strings to match. I think we should use capitalization whenever the message directly references something in the AWS console. For example, here it should be 'Auto Scaling Group' as capitalized.

tpetchel: I think we should call it whatever Amazon calls it. For example, "Auto Scaling group".
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html

randomcamel:
Looks great! Let's chat briefly with folks tomorrow to identify a standard for us to use going forward, then merge this.

lib/chef/provider/aws_auto_scaling_group.rb
tyler-ball: Even the console is divided on this - sometimes I see `Group` and sometimes I see `group`.  But it looks like the `G` is only in titles and navigation bars, which typically use title casing anyways.  I'm declare it shall be `g`roup!

tyler-ball:
Thanks @jamescott !

PR #293, opened 2015-08-18
Adding api V2 for machine create and destroy, fixes #216

### LINKED ISSUE #216: Introduce the AWS SDK V2 to the code
We can depend on both APIs at the same time, so I think this is a 2 part story.  The first part is about exposing the new V2 client that allows users to specify the new [credentials](http://docs.aws.amazon.com/sdkforruby/api/index.html#Configuration) while also being backwards compatiable with old credentials.  This means we can write new resources (hello IAM roles) that leverage the V2 SDK while not having to convert all the existing resources in one giant PR.

- [ ] Pull in the new SDK and make sure there are no credentials issues
- [ ] Add driver/resource methods to expose the new client that don't conflict with the old client
- [ ] Add deprecation warnings (in code and README) that show which old credentials specifying methods will be deprecated in 2.0
- [ ] Add tests for the new and old clients that verify that all know credentials specifying methods work

It doesn't depend on Nokogiri!

tyler-ball:
First stab at adding in API V2 support.  It requires loading both libraries, loading different clients and setting up different credentials.  My idea is that we would keep `Credential2` around until we got rid of v2 and then we would just rename it credentials.  Thoughts?

\cc @jkeiser @fnichol @randomcamel 

Fixes #216

### DIFF
--- a/chef-provisioning-aws.gemspec
+++ b/chef-provisioning-aws.gemspec
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'chef', '>= 11.16.4'
   s.add_dependency 'chef-provisioning', '~> 1.3'
   s.add_dependency 'aws-sdk-v1', '>= 1.59.0'
+  s.add_dependency 'aws-sdk', '~> 2.1'
   s.add_dependency 'retryable', '~> 2.0.1'
   s.add_dependency 'ubuntu_ami', '~> 0.4.1'
 
--- a/lib/chef/provider/aws_instance.rb
+++ b/lib/chef/provider/aws_instance.rb
@@ -10,12 +10,19 @@ class Chef::Provider::AwsInstance < Chef::Provisioning::AWSDriver::AWSProvider
     message += " in VPC #{instance.vpc.id}" unless instance.vpc.nil?
     message += " in #{region}"
     converge_by message do
-      instance.delete
+      instance.terminate
     end
     converge_by "waited until instance #{new_resource} is :terminated" do
       # When purging, we must wait until the instance is fully terminated - thats the only way
       # to delete the network interface that I can see
-      wait_for_status(instance, :terminated, [AWS::EC2::Errors::InvalidInstanceID::NotFound])
+      instance.wait_until_terminated do |w|
+        # TODO look at `wait_for_status` - delay and max_attempts should be configurable
+        w.delay = 5
+        w.max_attempts = 60
+        w.before_wait do |attempts, response|
+          action_handler.report_progress "waited #{(attempts-1)*5}/#{60*5}s for #{instance.id} status to terminate..."
+        end
+      end
     end
   end
 end
--- a/lib/chef/provisioning/aws_driver/aws_resource.rb
+++ b/lib/chef/provisioning/aws_driver/aws_resource.rb
@@ -133,13 +133,14 @@ class AWSResource < Chef::Provisioning::AWSDriver::SuperLWRP
                         load_provider: true,
                         id: :name,
                         aws_id_prefix: nil)
+    # TODO less than Chef 12.4, this is necessary.  Above that, this is deprecated.
     self.resource_name = self.dsl_name
     @aws_sdk_class = sdk_class
     @aws_sdk_class_id = id
     @aws_id_prefix = aws_id_prefix
 
     # Go ahead and require the provider since we're here anyway ...
-    require "chef/provider/#{resource_name}" if load_provider
+    require "chef/provider/#{resource_name.to_s}" if load_provider
 
     option_name = :"#{resource_name[4..-1]}" if option_name == NOT_PASSED
     @aws_sdk_option_name = option_name
--- a/lib/chef/provisioning/aws_driver/aws_resource.rb
+++ b/lib/chef/provisioning/aws_driver/aws_resource.rb
@@ -133,14 +133,13 @@ class AWSResource < Chef::Provisioning::AWSDriver::SuperLWRP
                         load_provider: true,
                         id: :name,
                         aws_id_prefix: nil)
-    # TODO less than Chef 12.4, this is necessary.  Above that, this is deprecated.
-    self.resource_name = self.dsl_name
+    self.resource_name = convert_to_snake_case(self.name.split('::')[-1])
     @aws_sdk_class = sdk_class
     @aws_sdk_class_id = id
     @aws_id_prefix = aws_id_prefix
 
     # Go ahead and require the provider since we're here anyway ...
-    require "chef/provider/#{resource_name.to_s}" if load_provider
+    require "chef/provider/#{resource_name}" if load_provider
 
     option_name = :"#{resource_name[4..-1]}" if option_name == NOT_PASSED
     @aws_sdk_option_name = option_name
--- a/lib/chef/provisioning/aws_driver/credentials2.rb
+++ b/lib/chef/provisioning/aws_driver/credentials2.rb
@@ -1,0 +1,51 @@
+require "aws-sdk"
+require "aws-sdk-core/credentials"
+require "aws-sdk-core/shared_credentials"
+require "aws-sdk-core/instance_profile_credentials"
+require "aws-sdk-core/assume_role_credentials"
+
+class Chef
+module Provisioning
+module AWSDriver
+
+  class LoadCredentialsError < RuntimeError; end
+
+  # Loads the credentials for the AWS SDK V2
+  # Attempts to load credentials in the order specified at http://docs.aws.amazon.com/sdkforruby/api/index.html#Configuration
+  class Credentials2
+
+    attr_reader :profile_name
+
+    # @param [Hash] options
+    # @option options [String] :profile_name (ENV["AWS_DEFAULT_PROFILE"]) The profile name to use
+    #    when loading the config from '~/.aws/credentials'.  This can be nil.
+    def initialize(options = {})
+      @profile_name = options.fetch(:profile_name, nil) || ENV["AWS_DEFAULT_PROFILE"]
+    end
+
+    # Try to load the credentials from an ordered list of sources and return the first one that
+    # can be loaded successfully.
+    def get_credentials
+      shared_creds = ::Aws::SharedCredentials.new(:profile_name => profile_name)
+      instance_profile_creds = ::Aws::InstanceProfileCredentials.new(:retries => 1)
+
+      if ENV["AWS_ACCESS_KEY_ID"] && ENV["AWS_SECRET_ACCESS_KEY"]
+        creds = ::Aws::Credentials.new(
+          ENV["AWS_ACCESS_KEY_ID"],
+          ENV["AWS_SECRET_ACCESS_KEY"],
+          ENV["AWS_SESSION_TOKEN"]
+        )
+      elsif shared_creds.set?
+        creds = shared_creds
+      elsif instance_profile_creds.set?
+        creds = instance_profile_creds
+      else
+        raise LoadCredentialsError.new("Could not load credentials from the environment variables, the .aws/credentials file or the metadata service")
+      end
+      creds
+    end
+  end
+
+end
+end
+end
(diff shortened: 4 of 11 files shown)


lib/chef/provisioning/aws_driver/aws_resource.rb
```diff
@@ -133,13 +133,14 @@ def self.aws_sdk_type(sdk_class,
                         load_provider: true,
                         id: :name,
                         aws_id_prefix: nil)
+    # TODO less than Chef 12.4, this is necessary.  Above that, this is deprecated.
     self.resource_name = self.dsl_name
```
randomcamel: Unrelated to this PR, but maybe we guard it with `if self.respond_to?(:resource_name)`?

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -1010,18 +1022,20 @@ def default_aws_keypair(action_handler, machine_spec)
...
       specs_and_servers = instances_for(specs_and_options.keys)
 
       by_bootstrap_options = {}
       specs_and_options.each do |machine_spec, machine_options|
-        actual_instance = specs_and_servers[machine_spec]
-        if actual_instance
-          if actual_instance.status == :terminated
-            Chef::Log.warn "Machine #{machine_spec.name} (#{actual_instance.id}) is terminated.  Recreating ..."
+        instance = specs_and_servers[machine_spec]
+        if instance
```
randomcamel: The `actual_instance` -> `instance` change makes it harder to see what's actually changed for the v2 driver.

randomcamel:
Looks good overall. Is the plan to thread some `if..then` into `driver.rb` to support v1 and v2 simultaneously?

tyler-ball:
> Looks good overall. Is the plan to thread some if..then into driver.rb to support v1 and v2 simultaneously?

Not right now.  I converted over `machine` to use all v2 so that we don't need to use `if...then`.  I think the same logic should work for converting on a resource-by-reource basis.

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -1034,7 +1058,7 @@ def create_servers(action_handler, specs_and_options, parallelizer, &block)
       end
 
       # Create the servers in parallel
-      parallelizer.parallelize(by_bootstrap_options) do |bootstrap_options, machine_specs|
+      parallelizer.parallel_do(by_bootstrap_options) do |bootstrap_options, machine_specs|
```
tyler-ball: I don't know why I had to do this, but machine batch wouldn't work if I didn't.  @jkeiser any help?

tyler-ball:
To get the tests passing I had to disable tag support for `aws_instance` (but not `machine`).  This is something I'll fix in the upcoming tagging module PR.

tyler-ball:
One idea I want to solicit feedback on - does it seem reasonable to keep converting these resources 1 by 1?  Do y'all see an issue with doing that?

If we do that, I expect we could get to the point of converting a few resources at a time and it would only take about .5 - 1 day to crank out each PR.

randomcamel:
Unless it requires a big ton of adapter code to go 1 by 1, I'd incline to that approach (and if you do 5 of them at a stretch, that's fine).

tyler-ball:
@randomcamel I don't think it will require adapter code.  The next resource I would tackle would be one of the `aws_` resources so we could make sure all the magic in the base class from `aws_resource.rb` (like `lookup_options`) works when some resources are using v1 clients and some are using v2 clients.

chef-provisioning-aws.gemspec
```diff
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'chef', '>= 11.16.4'
   s.add_dependency 'chef-provisioning', '~> 1.0'
   s.add_dependency 'aws-sdk-v1', '>= 1.59.0'
```
jkeiser: We probably shouldn't do this long run--if people are passing SDK v1 stuff into our APIs, their code will continue to work if we do this, but if we *remove* the dependency their code will break and they will know to fix it.

lib/chef/provider/aws_instance.rb
```diff
@@ -10,12 +10,19 @@ def destroy_aws_object(instance)
...
     converge_by "waited until instance #{new_resource} is :terminated" do
       # When purging, we must wait until the instance is fully terminated - thats the only way
       # to delete the network interface that I can see
-      wait_for_status(instance, :terminated, [AWS::EC2::Errors::InvalidInstanceID::NotFound])
+      instance.wait_until_terminated do |w|
+        # TODO look at `wait_for_status` - delay and max_attempts should be configurable
+        w.delay = 5
+        w.max_attempts = 60
+        w.before_wait do |attempts, response|
+          action_handler.report_progress "waited #{(attempts-1)*5}/#{60*5}s for #{instance.id} status to terminate..."
```
jkeiser: Is it useful to have a helper that handles this logic so it can be reused? Or do we not need that?

lib/chef/provisioning/aws_driver/aws_resource.rb
```diff
@@ -133,13 +133,13 @@ def self.aws_sdk_type(sdk_class,
                         load_provider: true,
                         id: :name,
                         aws_id_prefix: nil)
-    self.resource_name = self.dsl_name
+    self.resource_name = convert_to_snake_case(self.name.split('::')[-1])
```
jkeiser: 12.4 change?

lib/chef/provisioning/aws_driver/aws_resource.rb
```diff
@@ -133,13 +133,13 @@ def self.aws_sdk_type(sdk_class,
...
     @aws_sdk_class = sdk_class
     @aws_sdk_class_id = id
     @aws_id_prefix = aws_id_prefix
 
     # Go ahead and require the provider since we're here anyway ...
-    require "chef/provider/#{resource_name}" if load_provider
+    require "chef/provider/#{resource_name.to_s}" if load_provider
```
jkeiser: This should be unnecessary, .to_s is automatic when you embed something in a string.

lib/chef/provisioning/aws_driver/credentials2.rb
```diff
@@ -0,0 +1,51 @@
...
+require "aws-sdk-core/assume_role_credentials"
+
+class Chef
+module Provisioning
+module AWSDriver
+
+  class LoadCredentialsError < RuntimeError; end
+
+  # Loads the credentials for the AWS SDK V2
+  # Attempts to load credentials in the order specified at http://docs.aws.amazon.com/sdkforruby/api/index.html#Configuration
```
jkeiser: Yay doc pointer!

lib/chef/provisioning/aws_driver/credentials2.rb
```diff
@@ -0,0 +1,51 @@
...
+  class Credentials2
+
+    attr_reader :profile_name
+
+    # @param [Hash] options
+    # @option options [String] :profile_name (ENV["AWS_DEFAULT_PROFILE"]) The profile name to use
+    #    when loading the config from '~/.aws/credentials'.  This can be nil.
+    def initialize(options = {})
+      @profile_name = options.fetch(:profile_name, nil) || ENV["AWS_DEFAULT_PROFILE"]
```
jkeiser: Out of curiosity, why options.fetch and not options[:profile_name]?

lib/chef/provisioning/aws_driver/driver.rb
jkeiser: It's very strange that this change works at all.  `parallelize` returns a `ParallelEnumerable`, while `parallel_do` returns `nil`.

jkeiser: The main difference between them: `parallelize` returns a `ParallelEnumerable` *immediately*, and stuff happens in the background (but the `to_a` should wait until everything is done). `parallel_do` waits for everything to finish before it completes, but doesn't return anything.

lib/chef/resource/aws_instance.rb
```diff
@@ -1,10 +1,14 @@
 require 'chef/provisioning/aws_driver/aws_resource_with_entry'
 
 class Chef::Resource::AwsInstance < Chef::Provisioning::AWSDriver::AWSResourceWithEntry
+  # TODO when I try to set this as `Aws` instead of `AWS` I start getting lots of load errors
```
jkeiser: This type is used for creating the instance. I would be very wary of using v1 instance in one place and v2 in others; I think you need to sort this out before checking in. I'm happy to pair.

spec/integration/aws_tagged_items_spec.rb
```diff
@@ -84,24 +84,24 @@
...
-            aws_tags :project => 'FUN'
-          end
-        }.to update_an_aws_instance('test_instance'
-        ).and have_aws_instance_tags('test_instance',
-                                      { 'Name' => 'test_instance',
-                                        'project' => 'FUN'
-                                      }
-        ).and be_idempotent
-      end
+      # it "aws_instance 'test_instance' created with custom tag", :super_slow do
```
jkeiser: Why is this failing?

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: Yeah, I have no idea why I had to do this.  I'll try reverting and see if I can dig up the error message.

chef-provisioning-aws.gemspec
tyler-ball: What do you mean by `'this'`?  Keeping both dependencies, v1 and v2?  We have to do it now to support all the resources that _don't_ use v2.

lib/chef/provider/aws_instance.rb
tyler-ball: I want to make `delay` and `max_attempts` configurable, so I could see a helper to DRY that up.  But the actual messages we will display will be different for each thing we are waiting on.  I don't think its complicated enough to add a helper for formatting the messages.

lib/chef/provisioning/aws_driver/aws_resource.rb
tyler-ball: correct!

lib/chef/provisioning/aws_driver/aws_resource.rb
tyler-ball: I'm not sure why I did this...

lib/chef/provisioning/aws_driver/credentials2.rb
tyler-ball: ¯\_(ツ)_/¯

spec/integration/aws_tagged_items_spec.rb
tyler-ball: Because our tagging support is only setup to use the V1 tagging API.  I've got a PR for improving the tagging support (https://github.com/chef/chef-provisioning-aws/pull/299) that will allow me to solve this issue and un-comment this test.

lib/chef/resource/aws_instance.rb
tyler-ball: Got it figured out and fixed!  It was a load order issue :(

chef-provisioning-aws.gemspec
jkeiser: Yep, totally :) I see that is The Plan.

lib/chef/provider/aws_instance.rb
jkeiser: Okey doke!

lib/chef/provisioning/aws_driver/aws_resource.rb
jkeiser: So the issue is the use of dsl_name. OK, got it, sounds fine.

spec/integration/aws_tagged_items_spec.rb
randomcamel: Why comment it out instead of marking it `pending`?

lib/chef/provisioning/aws_driver/driver.rb
randomcamel: More specifically, why add what looks like a non-meaningful change into a bunch of meaningful changes?

tyler-ball: I don't like the word `actual` - I don't think it conveyed additional info.  I'm happy to revert this change if it makes the diff clearer and if it is just my OCD getting in the way :)

spec/integration/aws_tagged_items_spec.rb
tyler-ball: good point!

PR #119, opened 2015-03-01
RDS support

erikvanbrakel:
scratching another itch. I need to setup RDS instances in my current setup (actual app deployment is not with chef, so the DB should be there). Seeing as there's no support for this, I figured I'd give it a go ;-)

Work in progress!!

- [ ] DB instance
  - [x] Minimal implementation for create action
  - [ ] Add optional parameters
  - [ ] Implement delete action (don't forget param for creating a final snapshot or not!)
- [ ] DB subnet group
  - [x] Create action
  - [x] Delete action
  - [ ] Can it be modified?
- [ ] DB option groups
  - [x] Create action (combines creation and setting options)
  - [ ] Delete action
- [ ] DB parameter groups

Open source cookbook that has done this - https://github.com/gosuri/aws-rds-cookbook/blob/master/libraries/rds.rb

PR #294, opened 2015-08-18
Adding RDS Option Groups

tyler-ball:
Copied from https://github.com/chef/chef-provisioning-aws/pull/119

This needs to be completed by adding testing and making sure the resource/provider fit in with the existing rds resource styles.

### DIFF
--- a/lib/chef/provider/aws_rds_option_group.rb
+++ b/lib/chef/provider/aws_rds_option_group.rb
@@ -1,0 +1,45 @@
+require 'chef/resource/aws_resource'
+require 'chef/provider/aws_provider'
+
+class Chef::Provider::AwsRdsOptionGroup < Chef::Provider::AwsProvider
+
+  action :create do
+    fail "Can't create an option group without a description" if new_resource.description.nil?
+    fail "Can't create an option group without an engine name" if new_resource.engine_name.nil?
+    fail "Can't create an option group without a major engine version" if new_resource.major_engine_version.nil?
+
+    if existing_option_group == nil
+        creation_options = {
+            :option_group_name => new_resource.name,
+            :engine_name => new_resource.engine_name,
+            :major_engine_version => new_resource.major_engine_version,
+            :option_group_description => new_resource.description
+        }
+
+        converge_by "Creating new RDS option group" do
+            new_driver.rds.client.create_option_group(creation_options)
+            new_resource.save
+        end
+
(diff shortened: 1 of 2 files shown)


lib/chef/resource/aws_rds_option_group.rb
```diff
@@ -0,0 +1,25 @@
+require 'chef/resource/aws_resource'
+require 'chef/provisioning/aws_driver'
+
+class Chef::Resource::AwsRdsOptionGroup < Chef::Resource::AwsResource
+  self.resource_name = 'aws_rds_option_group'
```
tyler-ball: This should be updated to use the `aws_sdk_type` (even though the option groups don't have a model)

lib/chef/resource/aws_rds_option_group.rb
```diff
@@ -0,0 +1,25 @@
...
+  actions :create, :delete, :nothing
+  default_action :create
+
+  attribute :name, :kind_of => String, :name_attribute => true
+  attribute :description, :kind_of => String
+  attribute :engine_name, :kind_of => String
+  attribute :major_engine_version, :kind_of => String
+  attribute :options
+  attribute :apply_immediately, :default => false
+
```
tyler-ball: This needs the `aws_object` method, which can just return a hash for now

lib/chef/provider/aws_rds_option_group.rb
```diff
@@ -0,0 +1,45 @@
+require 'chef/resource/aws_resource'
+require 'chef/provider/aws_provider'
+
+class Chef::Provider::AwsRdsOptionGroup < Chef::Provider::AwsProvider
+
```
tyler-ball: This should be modified to use the `create_aws_object`, etc. methods

lib/chef/provider/aws_rds_option_group.rb
```diff
@@ -0,0 +1,45 @@
...
+  action :create do
+    fail "Can't create an option group without a description" if new_resource.description.nil?
+    fail "Can't create an option group without an engine name" if new_resource.engine_name.nil?
+    fail "Can't create an option group without a major engine version" if new_resource.major_engine_version.nil?
```
randomcamel: Why not mark these `required` in the resource?

tyler-ball:
Closing since this is a deprecated repo

PR #299, opened 2015-08-24
Tagging Refactor, fixes #281

### LINKED ISSUE #281: Better support for tags
Tag assignment through AWS occurs many ways.  Some tags need to be formatted differently, some can only be uploaded through the client, the `'Name'` tag is special, etc.

Lets introduce an `aws_taggable` module that can be applied to a resource.  This will introduce hook methods for reading tags, formatting them, updating them and deleting them.  Then the AwsProvider can leverage this logic and individual resources can overwrite the hooks.

We should also see if we can clean up the tagging duplication in the driver.rb

AC:  All resources available through chef-provisioning-aws should be taggable and documented how to do it.  Code duplication should be limited as much as possible.  Tagging should be tested for all resources.

tyler-ball:
The goal of this PR is to update the tagging implementation to allow

* Easier tagging for resource authors
  * A common pattern of how to tag resources that don't fit into the existing strategies
* Cleaner, tested, easer to maintain code
* For cookbook authors, a consistent interface for adding tags to resources
  * Specifying `aws_tags` as an attribute on the resource instead of in the `machine_options`
  * Requires https://github.com/chef/chef-provisioning/pull/424 if we want this

I introduce a new `AWSTagger` class that is responsible for the `converge_tags` method.  It will either succeed at this or raise an exception (and it has built in retry logic).  It expects to be given a strategy class that contains the client-specific details of how to interact with AWS and retrieve/update/delete tags.

Additional strategies will need to be written for ELB, CloudSearch, etc.

TODO:

- [x] Add unit tests for the new classes
- [x] Make sure integration tests still work
- [x] Get current feedback and see if this seems like a good idea
- [x] Add back tagging tests that were commented out in https://github.com/chef/chef-provisioning-aws/pull/293
- [ ] Update README
- [x] Make sure every `'Name'` tag has a retry around it
- [x] Finish machine_image and load_balancer


lib/chef/provider/aws_dhcp_options.rb
```diff
@@ -1,7 +1,21 @@
 require 'chef/provisioning/aws_driver/aws_provider'
+require 'chef/provisioning/aws_driver/tagging_strategy/ec2'
 require 'retryable'
 
 class Chef::Provider::AwsDhcpOptions < Chef::Provisioning::AWSDriver::AWSProvider
+
+  def converge_tags
```
tyler-ball: If we like this idea I'll DRY it up by (probably) putting it into a module.  But this is how a provider creates a strategy and the AWS Tagger instance.

lib/chef/provisioning/aws_driver/aws_taggable.rb
```diff
@@ -0,0 +1,18 @@
+module Chef::Provisioning::AWSDriver
+# This module is meant to be included in a resource that is taggable
+# This will add the appropriate attribute that can be converged by the provider
+# TODO it would be nice to not have two seperate modules (taggable/tagger)
+#   and just have the provider decorate the resource or vice versa.  Complicated
+#   by resources <-> providers being many-to-many.
+module AWSTaggable
```
tyler-ball: This module is simply for DRY purposes - if it seems unnecessary I can get rid of it

lib/chef/provisioning/aws_driver/aws_taggable.rb
```diff
@@ -0,0 +1,18 @@
+module Chef::Provisioning::AWSDriver
+# This module is meant to be included in a resource that is taggable
+# This will add the appropriate attribute that can be converged by the provider
+# TODO it would be nice to not have two seperate modules (taggable/tagger)
```
tyler-ball: I would like some feedback on this - any ideas for how I could have the resource automatically get a new attribute if I decorate the provider with the `converge_tags` method?

lib/chef/resource/aws_dhcp_options.rb
```diff
@@ -12,6 +12,7 @@
 #
 class Chef::Resource::AwsDhcpOptions < Chef::Provisioning::AWSDriver::AWSResourceWithEntry
   aws_sdk_type AWS::EC2::DHCPOptions
+  include Chef::Provisioning::AWSDriver::AWSTaggable
```
tyler-ball: This is what each resource that wants the `aws_tags` attribute will look like and include

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -38,6 +39,10 @@ class Driver < Chef::Provisioning::Driver
 
     attr_reader :aws_config
 
+    # We add the appropriate attributes to the base resources for tagging support
+    Chef::Resource::Machine.include(Chef::Provisioning::AWSDriver::AWSTaggable)
+    Chef::Provider::Machine.additional_machine_option_keys << :aws_tags
```
tyler-ball: This is where I leverage https://github.com/chef/chef-provisioning/pull/424 to declare that `machine` resources should include their `aws_tags` attributes inside the `machine_options` passed to the driver methods like `allocate_machine`.

lib/chef/provisioning/aws_driver/aws_tagger.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def converge_tags
+    if desired_tags.nil?
+      Chef::Log.debug "aws_tags not provided, nothing to converge"
+      return
+    end
+
+    # Duplication and normalization
+    n_desired_tags = Hash[desired_tags.map {|k,v| [k.to_s, v.to_s]}]
+    n_current_tags = Hash[current_tags.map {|k,v| [k.to_s, v.to_s]}]
```
randomcamel: Seems worth factoring out (and unit testing).

lib/chef/provisioning/aws_driver/aws_tagger.rb
```diff
@@ -0,0 +1,56 @@
...
+    # Tagging frequently fails so we retry with an exponential backoff, a maximum of 10 seconds
+    Retryable.retryable(:tries => 5, :sleep => lambda { |n| [2**n, 10].min }) do |retries, exception|
+      if retries > 0
+        Chef::Log.info "Retrying the tagging, previous try failed with #{exception.inspect}"
+      end
+      unless tags_to_set.empty?
+        action_handler.perform_action "creating tags #{tags_to_set}" do
+          set_tags(tags_to_set)
+        end
+        tags_to_set = []
```
randomcamel: Is there another way to break out of the `retryable` block?

lib/chef/provisioning/aws_driver/aws_taggable.rb
randomcamel: Seems perfectly reasonable that something marking itself `AWSTaggable` should thereby get `aws_tags`.

lib/chef/provisioning/aws_driver/aws_taggable.rb
randomcamel: IMO one of the confusing things about provisioning is the way it incompletely tries to blur the resource/provider distinction, so it seems fine to keep them separate and reduce the magic. If there's a way to somehow give clear feedback to the user when something is wrong--like they pass `aws_tags` into `machine_options` when the resource or provider isn't set up for tagging--then we keep to the way Chef resources are generally presented, and minimize the spooky-action-at-a-distance.

randomcamel:
Tests, Montresor!

lib/chef/provisioning/aws_driver/aws_taggable.rb
jkeiser: There are magic ways to cause the include in the Provider class to include stuff in the Resource class too, but honestly I don't think we'd want it--separate is fine. The pain we're seeing here is more that they *are* separate things--two separate classes both have to be modified to handle the noun and verb bit of this work. When we do get rid of providers, the separation pain will go away because there will be none.

jkeiser:
:+1:, sans one question: do you think there are worthwhile tests that can be added, or is this 100% backcompat and we think the tests already cover the things we might worry this patch will break?

lib/chef/provisioning/aws_driver/aws_taggable.rb
tyler-ball: :+1: Thanks!  I will keep the resource and provider modifications separate for now

lib/chef/provisioning/aws_driver/aws_tagger.rb
tyler-ball: What part is worth refactoring?  Changing all the keys/values into symbols?

For one-line changes like this I am okay without DRYing them up.  I think it can obfuscate the code to call a `keys_values_to_sym(desired_tags)` method here rather than just doing the operation inline.  Readers can quickly see what I am doing here where they have to go look up the definition of the `keys_values_to_sym` method if I refactor it.

lib/chef/provisioning/aws_driver/aws_tagger.rb
tyler-ball: I believe I can use the `break` keyword to break out of the `retryable` block.  But if I did that here then the `tags_to_delete` would not get ran.  Or are you suggesting I put them into 2 different retryable blocks?

tyler-ball:
I *definitely* am going to add and update tests.  I just wanted to get feedback on my proposed way of modifying a resource/provider to support tagging.  Since it seems that y'all were okay with that I'll go and finish updating this PR and adding tests before a final review.

lib/chef/provisioning/aws_driver/aws_tagger.rb
randomcamel: Fair enough.

lib/chef/provisioning/aws_driver/aws_tagger.rb
randomcamel: Nah, s'fine.

tyler-ball:
Closing this and splitting into 2 reviews - https://github.com/chef/chef-provisioning-aws/pull/314 and https://github.com/chef/chef-provisioning-aws/pull/315

PR #300, opened 2015-08-25
Add recursive_delete attribute to a aws_s3_bucket

stevendanna:
Non-empty S3 buckets have to be explicitly deleted with the #delete!
method rather than #delete.

### DIFF
--- a/lib/chef/provider/aws_s3_bucket.rb
+++ b/lib/chef/provider/aws_s3_bucket.rb
@@ -43,7 +43,11 @@ class Chef::Provider::AwsS3Bucket < Chef::Provisioning::AWSDriver::AWSProvider
 
   def destroy_aws_object(bucket)
     converge_by "delete S3 bucket #{new_resource.name}" do
-      bucket.delete
+      if new_resource.recursive_delete
+        bucket.delete!
+      else
+        bucket.delete
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


tyler-ball:
Hey @stevendanna I'm :+1: on this with 2 questions:

1.  What about the option of attempting a `delete` and if AWS throws a specific exception like 'bucket not empty' then we attempt a `delete!` (Just brainstorming out loud)?  I know we typically require users to specify `recursive true` on the `file` resource so by convention I'm :+1: on your current change, but I wanted to get your thoughts on that.  I don't have a good feel for how users want to manage S3 buckets to say which is better.
1. Can you add a test for this?  I know there isn't an `aws_s3_bucket_spec.rb` yet so I'm happy to help get this started.  But we have a 'no check in without tests' policy now that I'm trying to stick to.

stevendanna:
@tyler-ball 

Re (1), I was originally going to call `delete!` unconditionally, but I also don't know how people use this in general so I went with the attribute.  I don't have a strong opinion here.

Re (2), for sure.  I threw this over the wall just so I didn't forget while working on other stuff.  I'll write some tests for it tomorrow.

stevendanna:
@tyler-ball Tests added

tyler-ball:
Thanks @stevendanna !

PR #302, opened 2015-08-26
Renaming actual_instance because I don't think it provides more information than just instance

tyler-ball:
\cc @jkeiser @randomcamel 

Feel free to tell me to turn my OCD down a couple notches


randomcamel:
:shipit:

PR #303, opened 2015-08-26
Adding provides syntax to all providers to get rid of 12.4.0 chef warnings

tyler-ball:
\cc @jkeiser @randomcamel


randomcamel:
Still works with 12.2/12.3?

tyler-ball:
Yup!  Tested with 12.3.0 and 12.2.1 and ref_full.rb

PR #304, opened 2015-08-26
Add aws_internet_gateway provider, fixes #143

### LINKED ISSUE #143: `aws_internet_gateway` has no provider
`aws_internet_gateway` is defined as a resource but has no provider.

miguelcnf:
It has working create, update and destroy actions and can receive an options vpc attribute to attach with the internet gateway.

Integration tests included.

References:
http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Internet_Gateway.html
http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/EC2/InternetGateway.html

### DIFF
--- a/lib/chef/provider/aws_internet_gateway.rb
+++ b/lib/chef/provider/aws_internet_gateway.rb
@@ -1,0 +1,63 @@
+require 'chef/provisioning/aws_driver/aws_provider'
+require 'retryable'
+
+class Chef::Provider::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSProvider
+
+  protected
+
+  def create_aws_object
+    self.vpc = Chef::Resource::AwsVpc.get_aws_object(new_resource.vpc, resource: new_resource) if new_resource.vpc
+
+    converge_by "create internet gateway #{new_resource.name} in region #{region}" do
+      internet_gateway = new_resource.driver.ec2.internet_gateways.create
+      retry_with_backoff(AWS::EC2::Errors::InvalidInternetGatewayID::NotFound) do
+        internet_gateway.tags['Name'] = new_resource.name
+      end
+
+      if vpc
+        attach_vpc(vpc, internet_gateway)
+      end
+
+      internet_gateway
+    end
+  end
+
+  def update_aws_object(internet_gateway)
+    self.vpc = internet_gateway.vpc
+
+    if new_resource.vpc
+      desired_vpc = Chef::Resource::AwsVpc.get_aws_object(new_resource.vpc, resource: new_resource)
+      if vpc != desired_vpc
+        attach_vpc(desired_vpc, internet_gateway)
+      end
+    end
+  end
+
+  def destroy_aws_object(internet_gateway)
+    converge_by "delete internet gateway #{new_resource.name} in region #{region}" do
+      begin
+        detach_vpc(internet_gateway)
+        internet_gateway.delete
+      rescue AWS::EC2::Errors::InvalidInternetGatewayID::NotFound
+        raise "internet gateway #{internet_gateway.id} not found"
+      end
+    end
+  end
+
+  private
+
+  attr_accessor :vpc
+
+  def attach_vpc(vpc, internet_gateway)
+    action_handler.perform_action "attach vpc #{vpc.id} to #{internet_gateway.id}" do
+      internet_gateway.vpc = vpc
+    end
+  end
+
+  def detach_vpc(internet_gateway)
+    action_handler.perform_action "detach any currently attached vpc from internet gateway #{internet_gateway.id}" do
+      internet_gateway.detach(internet_gateway.vpc) if internet_gateway.vpc
+    end
+  end
+
+end
--- a/lib/chef/provider/aws_internet_gateway.rb
+++ b/lib/chef/provider/aws_internet_gateway.rb
@@ -6,13 +6,13 @@ class Chef::Provider::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSPro
 
   def action_detach
     internet_gateway = Chef::Resource::AwsInternetGateway.get_aws_object(new_resource.name, resource: new_resource)
-    detach_aws_object(internet_gateway)
+    detach_vpc(internet_gateway)
   end
 
   protected
 
   def create_aws_object
-    self.vpc = Chef::Resource::AwsVpc.get_aws_object(new_resource.vpc, resource: new_resource) if new_resource.vpc
+    desired_vpc = Chef::Resource::AwsVpc.get_aws_object(new_resource.vpc, resource: new_resource) if new_resource.vpc
 
     converge_by "create internet gateway #{new_resource.name} in region #{region}" do
       internet_gateway = new_resource.driver.ec2.internet_gateways.create
@@ -20,8 +20,8 @@ class Chef::Provider::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSPro
         internet_gateway.tags['Name'] = new_resource.name
       end
 
-      if vpc
-        attach_vpc(vpc, internet_gateway)
+      if desired_vpc
+        attach_vpc(desired_vpc, internet_gateway)
       end
 
       internet_gateway
@@ -29,11 +29,11 @@ class Chef::Provider::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSPro
   end
 
   def update_aws_object(internet_gateway)
-    self.vpc = internet_gateway.vpc
+    current_vpc = internet_gateway.vpc
 
     if new_resource.vpc
       desired_vpc = Chef::Resource::AwsVpc.get_aws_object(new_resource.vpc, resource: new_resource)
-      if vpc != desired_vpc
+      if current_vpc != desired_vpc
         attach_vpc(desired_vpc, internet_gateway)
       end
     end
@@ -41,22 +41,8 @@ class Chef::Provider::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSPro
 
   def destroy_aws_object(internet_gateway)
     converge_by "delete internet gateway #{new_resource.name} in region #{region}" do
-      begin
-        detach_aws_object(internet_gateway)
-        internet_gateway.delete
-      rescue AWS::EC2::Errors::InvalidInternetGatewayID::NotFound
-        raise "internet gateway #{internet_gateway.id} not found"
-      end
-    end
-  end
-
-  def detach_aws_object(internet_gateway)
-    converge_by "detach internet gateway #{new_resource.name} in region #{region}" do
-      begin
-        detach_vpc(internet_gateway)
-      rescue AWS::EC2::Errors::InvalidInternetGatewayID::NotFound
-        raise "internet gateway #{internet_gateway.id} not found"
-      end
+      detach_vpc(internet_gateway)
+      internet_gateway.delete
     end
   end
 
@@ -65,14 +51,16 @@ class Chef::Provider::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSPro
   attr_accessor :vpc
 
   def attach_vpc(vpc, internet_gateway)
-    action_handler.perform_action "attach vpc #{vpc.id} to #{internet_gateway.id}" do
+    converge_by "attach vpc #{vpc.id} to #{internet_gateway.id}" do
       internet_gateway.vpc = vpc
     end
   end
 
   def detach_vpc(internet_gateway)
-    action_handler.perform_action "detach any currently attached vpc from internet gateway #{internet_gateway.id}" do
-      internet_gateway.detach(internet_gateway.vpc) if internet_gateway.vpc
+    if internet_gateway.vpc
+      converge_by "detach vpc #{internet_gateway.vpc.id} from internet gateway #{internet_gateway.id}" do
+        internet_gateway.detach(internet_gateway.vpc)
+      end
     end
   end
 
--- a/lib/chef/provider/aws_internet_gateway.rb
+++ b/lib/chef/provider/aws_internet_gateway.rb
@@ -2,6 +2,12 @@ require 'chef/provisioning/aws_driver/aws_provider'
 require 'retryable'
 
 class Chef::Provider::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSProvider
+  provides :aws_internet_gateway
+
+  def action_detach
+    internet_gateway = Chef::Resource::AwsInternetGateway.get_aws_object(new_resource.name, resource: new_resource)
+    detach_aws_object(internet_gateway)
+  end
 
   protected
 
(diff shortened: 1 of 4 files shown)


miguelcnf:
Also it should solve #143.

tyler-ball:
@miguelcnf Right now a user can specify the `internet_gateway` attribute in the `aws_vpc` to have the VPC create/manage the internet gateway.  We decided on this because it didn't seem like an `internet_gateway` had any functionality apart from a VPC.  Did you have issues using this attribute, or is there a use case for a separate `internet_gateway` that we didn't think of?

joaogbcravo:
@tyler-ball One use case can be just the ability of tagging an internet gateway.

lib/chef/resource/aws_internet_gateway.rb
```diff
@@ -1,16 +1,43 @@
...
+# data bag `data/aws_internet_gateway/<name>`.
+#
+# API documentation for the AWS Ruby SDK for VPCs (and the object returned from `aws_object` can be found here:
+#
+# - http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/EC2/InternetGateway.html
+#
+class Chef::Resource::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSResourceWithEntry
+  aws_sdk_type AWS::EC2::InternetGateway
 
+  require 'chef/resource/aws_vpc'
```
tyler-ball: You shouldn't need this require line - was it giving you an error if you didn't?

spec/integration/aws_internet_gateway_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+describe Chef::Resource::AwsInternetGateway do
+  extend AWSSupport
+
+  when_the_chef_12_server "exists", organization: 'foo', server_scope: :context do
+    with_aws "with a VPC and an internet gateway" do
+      vpc = nil
+      internet_gateway = nil
+
+      before {
```
tyler-ball: Instead of this before block you should use

```ruby
aws_vpc "test_internet_gateway" do
  cidr_block "10.0.0.0/24"
end

# Later when you need to reference it...
test_internet_gateway.aws_object
```

Then VPC will get automatically torn down at the end then

spec/integration/aws_internet_gateway_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        }.to create_an_aws_internet_gateway('test_internet_gateway').and be_idempotent
+      end
+
+      it "aws_internet_gateway 'test_internet_gateway' with attached vpc" do
+        expect_recipe {
+          aws_internet_gateway 'test_internet_gateway' do
+            vpc vpc.id
+          end
+        }.to create_an_aws_internet_gateway('test_internet_gateway').and be_idempotent
+        filters = [
```
tyler-ball: Instead of these filters you should be able to specify
```ruby
}.to create_an_aws_internet_gateway('test_internet_gateway', vpc: test_internet_gateway.aws_object).and be_idempotent

File: lib/chef/resource/aws_internet_gateway.rb [hunk 4]
```diff
@@ -1,16 +1,43 @@
...
 
+  #
+  # A vpc to attach to the internet gateway.
+  #
+  # May be one of:
+  # - The name of an `aws_vpc` Chef resource.
+  # - An actual `aws_vpc` resource.
+  # - An AWS `VPC` object.
+  #
+  attribute :vpc, kind_of: [ String, AwsVpc, AWS::EC2::VPC ]
```
tyler-ball: Will you update this to take a `FalseClass` as well, and in the provider detach the internet gateway from the VPC if this attribute is false?

tyler-ball: Actually, I take it back - I think instead of `FalseClass` here you should just add a `:detach` action that detaches it from the VPC.  The `:destroy` should then just leverage the `action_detach` method.

tyler-ball:
Hey @miguelcnf thanks for this PR - overall I think it is _excellent_.  I had a couple of minor comments and 1 more request before we accept it.

The VPC has [existing logic](https://github.com/chef/chef-provisioning-aws/blob/79a51d4396fb996d50548f885eaa4dceff6e5f0d/lib/chef/provider/aws_vpc.rb#L149-L194) for managing an Internet Gateway.  Please update it to leverage this new provider by changing that method to look something like

```ruby
  def update_internet_gateway(vpc)
    current_ig = vpc.internet_gateway
    case new_resource.internet_gateway
    when String, Chef::Resource::AwsInternetGateway, AWS::EC2::InternetGateway
      new_ig = Chef::Resource::AwsInternetGateway.get_aws_object(new_resource.internet_gateway, resource: new_resource)
      if !current_ig
        aws_internet_gateway new_ig do
          vpc vpc
        end
      elsif current_ig != new_ig
        aws_internet_gateway current_ig do
          if current_ig.tags['OwnedByVPC'] == vpc.id
            action :destroy
          else
            action :detach
          end
        end
      end
    when true
      if !current_ig
        aws_internet_gateway "VPC managed IG" do
          vpc vpc
          aws_tags 'OwnedByVPC' => vpc.id
        end
      end
    when false
      aws_internet_gateway current_ig do
        if current_ig.tags['OwnedByVPC'] == vpc.id
          action :destroy
        else
          action :detach
        end
      end
    end
  end
```

Then will you update the Internet Gatewa tests to get full coverage of the possible scenarios?  Thank you!

miguelcnf:
Hey @tyler-ball thanks for the feedback - I'll try and pick the changes up later today.

spec/integration/aws_internet_gateway_spec.rb
miguelcnf: I've found that with that approach if an unexpected error happens which causes rspec to fail - the vpc (or really any other supporting resource created so far) wouldn't get destroyed - hence the "manual" before/after approach.

I understand the value/easiness of reusing chef resources in the specs but personally I'd like to avoid having to manually check/destroy test resources in case of unexpected failure of the test framework.

I'll change it if you think it'd better anyway - let me know.

spec/integration/aws_internet_gateway_spec.rb
miguelcnf: I thought those matchers would only validate that the aws_internet_gateway would be called with those attributes and not check if it had really attached the vpc/gw ?

miguelcnf:
@tyler-ball I think I've covered most of the suggestions.

Let me know if you think we need something else.

spec/integration/aws_internet_gateway_spec.rb
tyler-ball: Hmm, it should be destroying the resource - do you get any messages about `failure during after block` or something like that from RSpec?  [This](https://github.com/chef/chef-provisioning-aws/blob/master/spec/aws_support.rb#L138-L156) is the logic for setting up the resource for create/destroy in the before/after blocks.

lib/chef/resource/aws_internet_gateway.rb
```diff
@@ -1,16 +1,46 @@
...
+  aws_sdk_type AWS::EC2::InternetGateway
 
-class Chef::Resource::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSResource
-  aws_sdk_type AWS::EC2::InternetGateway, load_provider: false, id: :id
+  require 'chef/resource/aws_vpc'
 
+  #
+  # Extend actions for the internet gateway
+  #
+  actions :create, :destroy, :detach, :purge, :nothing
```
tyler-ball: You don't need `:purge` or `:nothing`

lib/chef/provider/aws_vpc.rb
```diff
@@ -148,47 +149,36 @@ def update_vpc_attributes(vpc)
...
-          current_ig = AWS.ec2(config: vpc.config).internet_gateways.create
-          retry_with_backoff(NeverObtainedExistence) do
-            raise NeverObtainedExistence.new("Internet gateway for VPC #{vpc.id} never obtained existence") unless current_ig.exists?
+        elsif current_ig != new_ig
+          aws_internet_gateway current_ig do
+            if current_ig.tags['OwnedByVPC'] == vpc.id
+              action :destroy
+            else
+              action :detach
+            end
```
tyler-ball: It looks like thei logic here is missing attaching the `new_ig`.  I think you could make lines 154-166 look like

```ruby
if current_ig != new_ig
  if current_ig
    aws_internet_gateway current_ig do
      if current_ig.tags['OwnedByVPC'] == vpc.id
        action :destroy
      else
        action :detach
      end
    end
  end
  aws_internet_gateway new_ig do
    vpc vpc.id
  end
end
```

lib/chef/provider/aws_internet_gateway.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  attr_accessor :vpc
+
+  def attach_vpc(vpc, internet_gateway)
+    action_handler.perform_action "attach vpc #{vpc.id} to #{internet_gateway.id}" do
+      internet_gateway.vpc = vpc
+    end
+  end
+
+  def detach_vpc(internet_gateway)
```
tyler-ball: Because this method is only called once, you can inline it.  And because its already called in the scope of a `converge_by` block you do not need the `action_handler.perform_action`.

lib/chef/provider/aws_internet_gateway.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+    end
+  end
+
+  def destroy_aws_object(internet_gateway)
+    converge_by "delete internet gateway #{new_resource.name} in region #{region}" do
+      begin
+        detach_aws_object(internet_gateway)
+        internet_gateway.delete
+      rescue AWS::EC2::Errors::InvalidInternetGatewayID::NotFound
```
tyler-ball: Don't bother rescuing their error and re-raising our own - their error is informative enough.

lib/chef/provider/aws_internet_gateway.rb
```diff
@@ -0,0 +1,79 @@
...
+        raise "internet gateway #{internet_gateway.id} not found"
+      end
+    end
+  end
+
+  def detach_aws_object(internet_gateway)
+    converge_by "detach internet gateway #{new_resource.name} in region #{region}" do
+      begin
+        detach_vpc(internet_gateway)
+      rescue AWS::EC2::Errors::InvalidInternetGatewayID::NotFound
```
tyler-ball: Ditto to https://github.com/chef/chef-provisioning-aws/pull/304/files#r39115479

spec/integration/aws_internet_gateway_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'spec_helper'
```
tyler-ball: Will you add a spec that tests the scenario of switching an internet gateway from VPC1 to VPC2?  This should test the `update_aws_object` method

lib/chef/provider/aws_internet_gateway.rb
```diff
@@ -0,0 +1,79 @@
+require 'chef/provisioning/aws_driver/aws_provider'
```
tyler-ball: I realized I was making a bunch of nit-picky comments here.  I've left them in the file for reference, but I rewrote this according to what I think fits our style guidelines better.  I removed some of the unecessary DRYing and made the actions more straightforward by putting all the `converge_by` logic into them.

```ruby
require 'chef/provisioning/aws_driver/aws_provider'
require 'retryable'

class Chef::Provider::AwsInternetGateway < Chef::Provisioning::AWSDriver::AWSProvider
  provides :aws_internet_gateway

  def action_detach
    internet_gateway = Chef::Resource::AwsInternetGateway.get_aws_object(new_resource.name, resource: new_resource)
    detach_vpc(internet_gateway)
  end

  protected

  def create_aws_object
    desired_vpc = Chef::Resource::AwsVpc.get_aws_object(new_resource.vpc, resource: new_resource) if new_resource.vpc

    converge_by "create internet gateway #{new_resource.name} in region #{region}" do
      internet_gateway = new_resource.driver.ec2.internet_gateways.create
      retry_with_backoff(AWS::EC2::Errors::InvalidInternetGatewayID::NotFound) do
        internet_gateway.tags['Name'] = new_resource.name
      end
    end

    if desired_vpc
      attach_vpc(desired_vpc, internet_gateway)
    end

    internet_gateway
  end

  def update_aws_object(internet_gateway)
    current_vpc = internet_gateway.vpc

    if new_resource.vpc
      desired_vpc = Chef::Resource::AwsVpc.get_aws_object(new_resource.vpc, resource: new_resource)
      if current_vpc != desired_vpc
        # TODO do we need a detach here first?
        attach_vpc(desired_vpc, internet_gateway)
      end
    end
  end

  def destroy_aws_object(internet_gateway)
    detach_vpc(internet_gateway)
    converge_by "delete internet gateway #{new_resource.name} in region #{region}" do
      internet_gateway.delete
    end
  end

  private

  def attach_vpc(vpc, internet_gateway)
    converge_by "attach vpc #{vpc.id} to #{internet_gateway.id}" do
      internet_gateway.vpc = vpc
    end
  end

  def detach_vpc(internet_gateway)
    if internet_gateway.vpc
      converge_by "detach VPC #{internet_gateway.vpc.id} from internet gateway #{internet_gateway.id}" do
        internet_gateway.detach(internet_gateway.vpc)
      end
    end
  end

end
```

spec/integration/aws_internet_gateway_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      context 'add an internet gateway and attach a vpc' do
+        it "aws_internet_gateway 'test_internet_gateway' attach vpc" do
+          expect_recipe {
+            aws_internet_gateway 'test_internet_gateway' do
+              vpc test_vpc_igw.aws_object.id
+            end
+          }.to create_an_aws_internet_gateway('test_internet_gateway',
+                                              vpc: test_vpc_igw.aws_object).and be_idempotent
+
+          expect(test_vpc_igw.aws_object.internet_gateway.id).not_to be_nil
```
tyler-ball: This expect is unnecessary - if the internet_gateway says its attached I think we can trust AWS

spec/integration/aws_internet_gateway_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+      end
+
+      context 'detach an internet gateway from a vpc' do
+        it "aws_internet_gateway 'test_internet_gateway' detach vpc" do
```
tyler-ball: This test is actually testing the `destroy` logic (which we need to test).  Can you add another test which tests that the `:detach` action detaches the IG from the VPC without deleting it?

tyler-ball:
Hey @miguelcnf this is looking really great.  The VPC looks muuuuuuuch cleaner.

I know I made a bunch of comments but they are mostly style comments, which I think are only necessary because we haven't published a style guide.  Other than that, it's just test coverage.  This is a really awesome PR and it's super close to being merged.  If you need help getting it across the finish line, just let us know!

tyler-ball:
One other question - did you run the VPC tests to make sure they still work?

miguelcnf:
Hey @tyler-ball, I haven't got back to you on this cause I was offline for a 2 weeks vacation.

I'll give it a look and try and update this today.

lib/chef/provider/aws_internet_gateway.rb
miguelcnf: No need to detach when updating with an already (different) attached vpc - from the docs:

> Attaches this internet gateway to the given VPC. If this gateway is already attached to a different VPC, it will be detached from that one first. If you pass nil, then this internet gateway will

miguelcnf:
@tyler-ball I think I've covered most of the suggestions. Have a look and let me know.

lib/chef/provider/aws_internet_gateway.rb
tyler-ball: awesome!

tyler-ball:
Merged to master in 70c3f4e877c9b8efb3e4c89e516c49413a84ee6b

Thanks @miguelcnf !

PR #305, opened 2015-08-31
Add aws_vpc_peering_connection.

joaogbcravo:
Added 'create', 'update', and 'destroy' actions for managing vpc peering connections.
Can accept an option to automatically accept a peering connection when the user is the owner of both vpc's.

Updated aws_vpc provider to destroy vpc_peering_requests when vpc is destroyed.
Updated aws_vpc resourse to allow optional names (vpc_id and peer_vpc_id)

Integration tests included.

### DIFF
--- a/lib/chef/provider/aws_vpc.rb
+++ b/lib/chef/provider/aws_vpc.rb
@@ -101,6 +101,31 @@ class Chef::Provider::AwsVpc < Chef::Provisioning::AWSDriver::AWSProvider
           end
         end
       end
+
+      #SDK V2
+      vpc_peering_connections = []
+      %w(
+        requester-vpc-info.vpc-id
+        accepter-vpc-info.vpc-id
+      ).each do |filter|
+        vpc_peering_connections += new_resource.driver.ec2_client.describe_vpc_peering_connections({
+            :filters => [
+                {
+                    :name => filter,
+                    :values => [vpc.id]
+                }
+            ]
+        }).vpc_peering_connections
+      end
+
+      vpc_peering_connections.each do |pc_type|
+        pc_resource = new_resource.driver.ec2_resource.vpc_peering_connection(pc_type.vpc_peering_connection_id)
+        Cheffish.inline_resource(self, action) do
+          aws_vpc_peering_connection pc_resource do
+            action :purge
+          end
+        end
+      end
     end
 
     # Detach or destroy the internet gateway
--- a/lib/chef/provider/aws_vpc_peering_connection.rb
+++ b/lib/chef/provider/aws_vpc_peering_connection.rb
@@ -1,0 +1,88 @@
+require 'chef/provisioning/aws_driver/aws_provider'
+require 'retryable'
+
+class Chef::Provider::AwsVpcPeeringConnection < Chef::Provisioning::AWSDriver::AWSProvider
+  provides :aws_vpc_peering_connection
+
+  def action_create
+    vpc_peering_connection = super
+    accept_connection(vpc_peering_connection, new_resource)
+  end
+
+  def action_accept
+    existing_vpc_peering_connection = new_resource.aws_object
+    accept_connection(existing_vpc_peering_connection, new_resource)
+  end
+
+  protected
+
+  def create_aws_object
+    if new_resource.vpc.nil?
+      raise "VCP peering connection create action for '#{new_resource.name}' requires the 'vpc' attribute."
+    elsif new_resource.peer_vpc.nil?
+      raise "VCP peering connection create action for '#{new_resource.name}' requires the 'peer_vpc' attribute."
+    end
+
+    options = {}
+    options[:vpc_id] = new_resource.vpc
+    options[:peer_vpc_id] = new_resource.peer_vpc
+    options[:peer_owner_id] = new_resource.peer_owner_id unless new_resource.peer_owner_id.nil?
+    options = AWSResource.lookup_options(options, resource: new_resource)
+
+    ec2_resource = new_resource.driver.ec2_resource
+    vpc = ec2_resource.vpc(options[:vpc_id])
+
+    converge_by "create peering connection #{new_resource.name} in VPC #{new_resource.vpc} (#{vpc.id}) and region #{region}" do
+      vpc_peering_connection = vpc.request_vpc_peering_connection(options)
+
+      retry_with_backoff(Aws::EC2::Errors::ServiceError) do
+        ec2_resource.create_tags({
+          :resources => [vpc_peering_connection.id],
+          :tags => [
+            {
+              :key => "Name",
+              :value => new_resource.name
+            }
+          ]
+        })
+      end
+      vpc_peering_connection
+    end
+  end
+
+  def update_aws_object(vpc_peering_connection)
+    vpc_id = vpc_peering_connection.requester_vpc_info.vpc_id
+    peer_vpc_id = vpc_peering_connection.accepter_vpc_info.vpc_id
+    peer_owner_id = vpc_peering_connection.accepter_vpc_info.owner_id
+
+    desired_vpc_id = Chef::Resource::AwsVpc.get_aws_object_id(new_resource.vpc, resource: new_resource)
+    desired_peer_vpc_id = Chef::Resource::AwsVpc.get_aws_object_id(new_resource.peer_vpc, resource: new_resource)
+    desired_peer_owner_id = new_resource.peer_owner_id
+
+    if desired_vpc_id && vpc_id != desired_vpc_id
+      raise "VCP peering connection requester vpc cannot be changed after being created! Desired requester vpc id for #{new_resource.name} (#{vpc_peering_connection.id}) was \"#{desired_vpc_id}\" and actual id is \"#{vpc_id}\""
+    end
+    if desired_peer_vpc_id && peer_vpc_id != desired_peer_vpc_id
+      raise "VCP peering connection accepter vpc cannot be changed after being created! Desired accepter vpc id for #{new_resource.name} (#{vpc_peering_connection.id}) was \"#{desired_peer_vpc_id}\" and actual id is \"#{peer_vpc_id}\""
+    end
+    if desired_peer_owner_id && peer_owner_id != desired_peer_owner_id
+      raise "VCP peering connection accepter owner id vpc cannot be changed after being created! Desired accepter vpc owner id for #{new_resource.name} (#{vpc_peering_connection.id}) was \"#{desired_peer_owner_id}\" and actual owner id is \"#{peer_owner_id}\""
+    end
+  end
+
+  def destroy_aws_object(vpc_peering_connection)
+    converge_by "delete #{new_resource.to_s} in #{region}" do
+      unless ['deleted', 'failed', 'deleting'].include? vpc_peering_connection.status.code
+        vpc_peering_connection.delete
+      end
+    end
+  end
+
+  private
+
+  def accept_connection(vpc_peering_connection, new_resource)
+    if new_resource.peer_owner_id.nil? or new_resource.peer_owner_id == new_resource.driver.account_id
+      vpc_peering_connection.accept
+    end
+  end
+end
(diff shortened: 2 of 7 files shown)


lib/chef/resource/aws_vpc_peering_connection.rb
```diff
@@ -0,0 +1,77 @@
...
+  # The target VPC account id to peer
+  #
+  # If not specified, will be assumed that the target VPC belongs to the current account.
+  #
+  attribute :peer_owner_id, kind_of: String, required: false
+
+  #
+  # If you are the owner of the target_vpc, you can accept the peering connection automatically.
+  #
+  attribute :automatic_accept, equal_to: [ true, false ], required: false, default: false
```
tyler-ball: What happens if `automatic_accept` is left off?  Does the peer need to make an API call to accept it?

lib/chef/resource/aws_vpc_peering_connection.rb
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  def aws_object
+    driver, id = get_driver_and_id
+    result = driver.ec2_resource.vpc_peering_connection(id) if id
+
+    begin
+      # try accessing it to find out if it exists
+      result.requester_vpc if result
+    rescue Aws::EC2::Errors::ServiceError
```
tyler-ball: What are some examples of these errors and how they occur?  I would hope that if we search by id and get a result back then this wouldn't be possible.  Will it return declined, expired or deleted results?

tyler-ball: The [possible values](http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Types/VpcPeeringConnectionStateReason.html) of status.code are `initiating-request, pending-acceptance, active, deleted, rejected, failed, expired, provisioning, deleting`.  It may be better to check if the `status.code` is one of `initiating-request, pending-acceptance, active, provisioning` rather than calling `requester_vpc` but I am up to discussion on this.

lib/chef/provider/aws_vpc_peering_connection.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def create_aws_object
+    options = {}
+    options[:vpc_id] = new_resource.vpc
+    options[:peer_vpc_id] = new_resource.peer_vpc
+    options[:peer_owner_id] = new_resource.peer_owner_id unless new_resource.peer_owner_id.nil?
+    options = AWSResource.lookup_options(options, resource: new_resource)
+
+    ec2_resource = new_resource.driver.ec2_resource
+    self.vpc = ec2_resource.vpc(options[:vpc_id])
```
tyler-ball: Why are you setting `self.vpc` on the provider?

lib/chef/provider/aws_vpc_peering_connection.rb
```diff
@@ -0,0 +1,89 @@
...
+  protected
+
+  def create_aws_object
+    options = {}
+    options[:vpc_id] = new_resource.vpc
+    options[:peer_vpc_id] = new_resource.peer_vpc
+    options[:peer_owner_id] = new_resource.peer_owner_id unless new_resource.peer_owner_id.nil?
+    options = AWSResource.lookup_options(options, resource: new_resource)
+
+    ec2_resource = new_resource.driver.ec2_resource
```
tyler-ball: I would like your opinion - how do you feel about the `ec2_client` and `ec2_resource` method names?

lib/chef/provider/aws_vpc_peering_connection.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def update_status(vpc_peering_connection, new_resource)
+    if new_resource.automatic_accept
+      if (new_resource.peer_owner_id.nil? || new_resource.peer_owner_id == new_resource.driver.account_id)
+        vpc_peering_connection.accept
+      end
+    end
+  end
+
+  attr_accessor :vpc
```
tyler-ball: We typically put these at the top of the class - would you move this there?

lib/chef/provider/aws_vpc_peering_connection.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def destroy_aws_object(vpc_peering_connection)
+    converge_by "delete #{new_resource.to_s} in #{region}" do
+      begin
+        unless ['deleted', 'failed', 'deleting'].include? vpc_peering_connection.status.code
+          vpc_peering_connection.delete
+        end
+      rescue Aws::EC2::Errors
+        raise "#{new_resource.to_s} could not be deleted."
```
tyler-ball: I would rather just let the original error bubble up

tyler-ball:
@joaogbcravo Thank you for this PR.  This is AWESOME work!  I had a few minor questions / comments, but overall this is :star: :rocket: :cake:!  AND it has test coverage so :100: :tada: :gem: for you as well.  

I think the biggest thing I would like your feedback on is how you see these working between different accounts.  Would account A and account B have to both run the same chef-provisioning resource with the acceptor having `automatic_accept` set to true?  Thats how I see it working.  They could also accept it through the console or other non-provisioning methods, but that doesn't sound fun :smile:

lib/chef/resource/aws_vpc_peering_connection.rb
joaogbcravo: Well, I thought there are no good reasons to automatic accept a peering account, unless you haven't permissions too, and in that case, no solution.

For the ones that have permissions, but at first set the "automatic_accept" to false (or let it empty), they always can update their recipe, updating the attribute to true, and then on the "update" action,  the peer connection is accepted.

Thinking more about this...I don't think it's necessary an action just to accept the request. And I don't think it make any sense at all create an action to reject some peer connection. What do you think?

lib/chef/resource/aws_vpc_peering_connection.rb
joaogbcravo: Yup, I guess you are right about it.  I will update it.

lib/chef/provider/aws_vpc_peering_connection.rb
joaogbcravo: Actually I guess there are no need to this "attr_accessor"

lib/chef/provider/aws_vpc_peering_connection.rb
joaogbcravo: Hum.. I tried to follow the same "coding structure" of route table provider. I haven't strong opinions about this, so whatever you think is best to maintain.

lib/chef/provider/aws_vpc_peering_connection.rb
joaogbcravo: Not sure what do you want to know about my feelings, I guess we can chat in gitter :)

lib/chef/resource/aws_vpc_peering_connection.rb
joaogbcravo: Actually, even when the states.code is "failed" or "expired", the peering connection was created, so the aws_object should return it anyway right?

lib/chef/resource/aws_vpc_peering_connection.rb
joaogbcravo: Probably it's need an action to just accept a request, see the overall comment.

joaogbcravo:
Thanks @tyler-ball :)
I couldn't test with different accounts. The account that makes the request (A) don't need to set automatic_accept because can't accept it (because it's not the owner of b). The account that accepts the request (B), can't "automatic_accept" it, because it's not the owner of A. I guess, we would need an action to accept peer connections. 

Now, that probably makes sense to create an "accept" action, would we need a "reject" action?

lib/chef/resource/aws_vpc_peering_connection.rb
tyler-ball: Okay, I just checked a bunch of stuff.  By convention `aws_object` always returns the aws object, regardless of its status.  So even if the peering connection is `deleting` or another 'invalid' status we should return it.

So I think your current code is correct!  Sorry for confusing this issue with my original question.  I was mistaken thinking that we should return nil if the status bad.  That requirement is on the provider and other callers of `aws_object`.

I also filed https://github.com/aws/aws-sdk-ruby/issues/916 to see if we could get the `exists?` method added.

tyler-ball: Actually, one last note - I think we should catch the specific `Aws::EC2::Errors::InvalidVpcPeeringConnectionIDNotFound` exception rather than the more abstract `Aws::EC2::Errors::ServiceError` here.

lib/chef/provider/aws_vpc_peering_connection.rb
tyler-ball: Yup, let it bubble up!

tyler-ball:
From our gitter conversation - there should be 3 actions `[:create, :accept, :destroy]`.  `:accept` won't try to create the connection, it will only try to accept an existing one.  `:create` will accept the connection by default if the workstation account owns both VPCs and the `automatically_accept` attribute should be deleted.

The workflow would look something like:

1. Provisioning workstation 1 creates a peering request but cannot accept it because the peer VPC is owned by another account.  It stores the id `pcx-123456` into a data_bag (or other shared system).
2. Provisioning workstation 2 either does a search for `pending-acceptance` peer requests or knows the shared resource name.
3. Provisioning workstation 2 declares
```ruby
aws_vpc_peering_connection "shared resource name" do # or "pcx-123456"
  action :accept
end
```
to accept the request.

lib/chef/provider/aws_vpc.rb
```diff
@@ -99,6 +99,31 @@ def destroy_aws_object(vpc)
           end
         end
       end
+
+      #SDK V2
+      vpc_peering_connections = []
```
tyler-ball: Another testing request I forgot earlier - can you add a test to `aws_vpc_spec.rb` that tests this logic?  Something like...

```
context "foo" do
  aws_vpc "vpc 1"
  aws_vpc "vpc 2"
  aws_vpc_peering_connection "conn" do
    vpc "vpc 1"
    peer_vpc "vpc 2"
  end

  it "deletes the peer connection when deleting the vpc" do
    converge {
      aws_vpc "vpc 1" do
        action :destroy
      end
    }
    # Lookup the peering connections
    expect(peering_connection).to eq(nil)
  end
end
```

lib/chef/resource/aws_vpc_peering_connection.rb
```diff
@@ -0,0 +1,73 @@
...
+  # The Local VPC to peer
+  #
+  # May be one of:
+  # - The name of an `aws_vpc` Chef resource.
+  # - An actual `aws_vpc` resource.
+  # - An AWS `VPC` object.
+  #
+  # This is required for new peering connections.
+  #
+  attribute :vpc, kind_of: [ String, AwsVpc, AWS::EC2::VPC ], required: true
```
tyler-ball: You should remove `required: true` from `vpc` and `vpc_peer` so that someone trying to accept a connection could declare

```ruby
# They could also use the id instead of the resource name
aws_vpc_peering_connection "conn" do
  action: accept
end
```

tyler-ball:
A few minor notes/updates, but this is super close to getting merged.  It'll be released as part of 1.4.0.  Thanks for your continued work!

spec/integration/aws_vpc_peering_connection_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+            :'accepter_vpc_info.vpc_id' => test_vpc_2.aws_object.id,
+            :'status.code' => 'failed'
+        ).and be_idempotent
+      end
+
+      it "aws_peering_connection 'test_peering_connection' with accept action, accepts a pending peering connection" do
+        pcx = nil
+        ec2_resource = driver.ec2_resource
+        expect_recipe {
+          ruby_block "upload s3 object" do
```
tyler-ball: `ruby_block "fetch VPC objects" do`

danieljimenez:
Nice! Any docs for this besides the spec?

joaogbcravo:
No, but it's pretty easy checking the resource and the examples in the integration tests ;) Any question let me know

PR #312, opened 2015-09-04
Update resources to allow route tables with pcx

joaogbcravo:
This PR depends on https://github.com/chef/chef-provisioning-aws/pull/305

### DIFF
--- a/lib/chef/provider/aws_route_table.rb
+++ b/lib/chef/provider/aws_route_table.rb
@@ -20,14 +20,23 @@ class Chef::Provider::AwsRouteTable < Chef::Provisioning::AWSDriver::AWSProvider
 
   def create_aws_object
     options = {}
-    options[:vpc] = new_resource.vpc
+    options[:vpc_id] = new_resource.vpc
     options = AWSResource.lookup_options(options, resource: new_resource)
-    self.vpc = Chef::Resource::AwsVpc.get_aws_object(options[:vpc], resource: new_resource)
+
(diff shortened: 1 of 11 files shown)


lib/chef/provider/aws_route_table.rb
```diff
@@ -18,14 +18,24 @@ def action_create
...
+      retry_with_backoff(Aws::EC2::Errors::ServiceError) do
+        route_table.create_tags({
+             :tags => [
+                 {
+                     :key => "Name",
+                     :value => new_resource.name
+                 }
+             ]
+         })
+        puts route_table.tags
```
tyler-ball: This debugging line can be removed

lib/chef/provider/aws_route_table.rb
```diff
@@ -61,8 +71,9 @@ def update_routes(vpc, route_table, ignore_route_targets = [])
     current_routes = {}
     route_table.routes.each do |route|
       # Ignore the automatic local route
-      next if route.target.id == 'local'
-      next if ignore_route_targets.find { |target| route.target.id.match(/#{target}/) }
+      route_target = route.gateway_id || route.instance_id || route.network_interface_id || route.vpc_peering_connection_id
```
tyler-ball: So route no longer has `route.target`.  Are we sure this is the full list of available targets?  Will this need to be updated in the future if new targets get added?

tyler-ball:
Specs look good - I had 2 questions then this is :+1: and ready for merge!

PR #314, opened 2015-09-09
Tagging Refactor Part 1, fixes #281

### LINKED ISSUE #281: Better support for tags

tyler-ball:
This replaces https://github.com/chef/chef-provisioning-aws/pull/299.  I split it into 2 reviews - this first one shows the architectural changes to enable the tagging refactor.

If a resource wants to use tagging it needs to add the `aws_tags` attribute.  For all the non-base resources this is done by including the `Chef::Provisioning::AWSDriver::AWSTaggable` module.  The base resources are modified by adding the module _and_ adding `aws_tags` to their  `additional_*_options` list.  This means the resource will get passed to the driver as part of the `*_options` hash passed to the hook methods.

For a resource to use tagging it needs to add the `converge_tags` method.  The base `AWSProvider` class will then call that method after converging the resource.  I introduced an `AWSTagger` class which has a `converge_tags` method that knows how to diff current tags from desired tags and converge them.  Because every AWS client handles this differently, I use a strategy pattern to supply the exact logic for each resource.

TODO
- [x] README & CHANGELOG updates

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,6 @@ gemfile
 gemspec
 
 #gem 'chef-provisioning', path: '../chef-provisioning'
-#gem 'chef-provisioning', github: 'chef/chef-provisioning', branch: 'master'
+gem 'chef-provisioning', github: 'chef/chef-provisioning', branch: 'master'
 #gem "pry-byebug"
 #gem "pry-stack_explorer"
--- a/lib/chef/provisioning/aws_driver/aws_resource.rb
+++ b/lib/chef/provisioning/aws_driver/aws_resource.rb
@@ -1,6 +1,8 @@
 require 'aws'
 require 'chef/provisioning/aws_driver/super_lwrp'
 require 'chef/provisioning/chef_managed_entry_store'
+# Enough resources will eventually require this that we put 1 require in here
+require 'chef/provisioning/aws_driver/aws_taggable'
 
 # Common AWS resource - contains metadata that all AWS resources will need
 module Chef::Provisioning::AWSDriver
@@ -65,6 +67,13 @@ class AWSResource < Chef::Provisioning::AWSDriver::SuperLWRP
   #
   def aws_object
     raise NotImplementedError, :aws_object
+  end
+
+  def aws_object_id
+    @aws_object_id ||= begin
+      o = aws_object
+      o.public_send(self.class.aws_sdk_class_id) if o
+    end
   end
 
   #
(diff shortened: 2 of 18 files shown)


Gemfile
```diff
@@ -4,6 +4,6 @@ gemfile
 gemspec
 
 #gem 'chef-provisioning', path: '../chef-provisioning'
-#gem 'chef-provisioning', github: 'chef/chef-provisioning', branch: 'master'
+gem 'chef-provisioning', github: 'chef/chef-provisioning', branch: 'master'
```
randomcamel: Intentional?

lib/chef/provisioning/aws_driver/aws_resource.rb
```diff
@@ -67,6 +69,13 @@ def aws_object
     raise NotImplementedError, :aws_object
   end
 
+  def aws_object_id
+    @aws_object_id ||= begin
+      o = aws_object
+      o.public_send(self.class.aws_sdk_class_id) if o
```
randomcamel: We do this `public_send` thing so often it seems like the classes should have an instance method that does it. (Not that I'm recommending it here.)

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -456,11 +452,13 @@ def ready_image(action_handler, image_spec, image_options)
       if actual_image.nil? || !actual_image.exists?
         raise 'Cannot ready an image that does not exist'
       else
+        image_options = (image_options || {}).to_h.dup
+        aws_tags = image_options.delete(:aws_tags) || {}
+        aws_tags['From-Instance'] = image_spec.reference['from-instance'] if image_spec.reference['from-instance']
```
randomcamel: Why the capitalization?

randomcamel:
:+1:

Gemfile
tyler-ball: Yes, because this depends on https://github.com/chef/chef-provisioning/pull/436

PR #323, opened 2015-09-15
Ensuring all V2 aws classes are prepended with :: to limit namespace scope

tyler-ball:
Fixes Delivery incident

\cc @cwebberOps @randomcamel


randomcamel:
:+1:

PR #326, opened 2015-09-18
specs for dynamic options in machine_batch, per chef-provisioning#438

redterror:
Two things here:
1. The `action` the existing spec was moved from `machine` to `machine_batch`.  The latter is in control (bug/feature?  not clear!) in this context, so I think this is just a bug in the test.
2. Add a spec for dynamic options used in `machine_batch`, which ought to also serve as an example of how to dynamically choose a subnet/az for a new node.

As mentioned in [chef-provisioning#438](https://github.com/chef/chef-provisioning/issues/438), I was hoping this spec would fail and shed some light on why it fails in my real recipe, but it passes.  Yay?

### DIFF
--- a/spec/integration/machine_batch_spec.rb
+++ b/spec/integration/machine_batch_spec.rb
@@ -12,16 +12,27 @@ describe Chef::Resource::MachineBatch do
 
       purge_all
       setup_public_vpc
+
+      azs = []
+      driver.ec2.availability_zones.each {|az| azs << az }
+      az = azs[1].name
+      aws_subnet 'test_subnet2' do
+        vpc 'test_vpc'
+        cidr_block '10.0.1.0/24'
+        availability_zone az
+        map_public_ip_on_launch true
+      end
+
       it "machine_batch creates multiple machines", :super_slow do
         expect_recipe {
           machine_batch 'test_machines' do
+            action :allocate
             (1..3).each do |i|
               machine "test_machine#{i}" do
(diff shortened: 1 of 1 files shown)


redterror:
Sorry I fail at markdown!

spec/aws_support.rb
```diff
@@ -44,7 +44,7 @@ def purge_all
 
   def setup_public_vpc
     aws_vpc 'test_vpc' do
-      cidr_block '10.0.0.0/24'
+      cidr_block '10.0.0.0/16'
```
tyler-ball: haha, I had to do the same thing in https://github.com/chef/chef-provisioning-aws/pull/346/files#diff-909db8c1e9f02815c26e8ebd3caa7b49R47

tyler-ball:
@redterror Just to clarify - this passes in a test but fails in a real recipe, correct?  Have you put a breakpoint in the test and looked at the actual machines in the AWS console to validate whether its the test that has a bug or the real recipe?

redterror:
That's correct, but I haven't put a breakpoint in the test.  I'm happy to do that and report back.

tyler-ball:
Hey @redterror will you rebase this off master?  I made some of the same changes in another PR that I merged.

redterror:
I'm late to the 1.5 party, but finally getting back to this...

tas50:
Sorry this took so long @redterror

PR #327, opened 2015-09-18
user_data needs to be base64 encoded in SDK V2, fixes #325

### LINKED ISSUE #325: Invalid BASE64 encoding of user data in 1.4.0
When upgrading to chef-provisioning{-aws} 1.4.0 from 1.3.1 I get the following error from AWS:

```Aws::EC2::Errors::InvalidParameterValue: Invalid BASE64 encoding of user data
Raw```

See https://gist.github.com/hh/6e8f0f7040f3eb0cf628

tyler-ball:
fixes #325

\cc @jkeiser @randomcamel @hh


hh:
works for me! Thanks!! :+1:

PR #330, opened 2015-09-20
removing non-base64 windows user_data

hh:


tyler-ball:
Nice catch - thanks @hh

hh:
hmm.... we may still have an issue (I was bootstrapping windows,nodes that already had that user_data executed...) as I'm not seeing user-data from the instances after my change

PR #332, opened 2015-09-21
Updating to support chef master

tyler-ball:
Chef 12.5.0 defines `Chef::NOT_PASSED` and chef-provisioning-aws was also defining this.  The double defining was causing issues with param validation.  There were also features (like `coerce` and `lazy_default`) being added to chef which are no longer required in 12.5

\cc @jkeiser @randomcamel

### DIFF
--- a/lib/chef/provisioning/aws_driver/aws_resource.rb
+++ b/lib/chef/provisioning/aws_driver/aws_resource.rb
@@ -5,199 +5,208 @@ require 'chef/provisioning/chef_managed_entry_store'
 require 'chef/provisioning/aws_driver/aws_taggable'
 
 # Common AWS resource - contains metadata that all AWS resources will need
-module Chef::Provisioning::AWSDriver
-class AWSResource < Chef::Provisioning::AWSDriver::SuperLWRP
-  actions :create, :destroy, :purge, :nothing
-  default_action :create
+class Chef
+module Provisioning
+module AWSDriver
 
-  def initialize(name, run_context=nil)
-    name = name.public_send(self.class.aws_sdk_class_id) if name.is_a?(self.class.aws_sdk_class)
-    super
-    if run_context
-      driver run_context.chef_provisioning.current_driver
-      chef_server run_context.cheffish.current_chef_server
-    end
-  end
+  class AWSResource < Chef::Provisioning::AWSDriver::SuperLWRP
+    actions :create, :destroy, :purge, :nothing
+    default_action :create
 
-  # Backwards compatibility for action :destroy
-  def action(*args)
-    if args == [ :delete ]
-      super(:destroy)
-    else
+    def initialize(name, run_context=nil)
+      name = name.public_send(self.class.aws_sdk_class_id) if name.is_a?(self.class.aws_sdk_class)
       super
-    end
-  end
-
-  # In Chef < 12.4 we need to define this method.  In > 12.4 this method is
-  # already defined for us so we don't need to overwrite it.
-  unless defined?(:action=)
-    def action=(value)
-      action(value)
-    end
-  end
-
-  #
-  # The desired driver.
-  #
-  attribute :driver, kind_of: Chef::Provisioning::Driver,
-                     coerce: (proc do |value|
-                               case value
-                               when nil, Chef::Provisioning::Driver
-                                 value
-                               else
-                                 run_context.chef_provisioning.driver_for(value)
-                               end
-                             end)
-
-  #
-  # The Chef server on which any IDs should be looked up.
-  #
-  attribute :chef_server, kind_of: Hash
-
-  #
-  # The managed entry store.
-  #
-  attribute :managed_entry_store, kind_of: Chef::Provisioning::ManagedEntryStore,
-                              lazy_default: proc { Chef::Provisioning::ChefManagedEntryStore.new(chef_server) }
-
-  #
-  # Get the current AWS object.  This should return the aws_object even if it has
-  # a status like 'deleted' or 'inactive'.  If there is an aws_object, we return it.
-  # Callers will need to check the status if they care.
-  #
-  def aws_object
-    raise NotImplementedError, :aws_object
-  end
-
-  def aws_object_id
-    @aws_object_id ||= begin
-      o = aws_object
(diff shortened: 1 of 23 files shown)


lib/chef/provisioning/aws_driver/super_lwrp.rb
```diff
@@ -7,30 +7,43 @@ class SuperLWRP < Chef::Resource::LWRPBase
...
-      if lazy_default || coerce
-        define_method(attr_name) do |arg=nil|
-          arg = instance_exec(arg, &coerce) if coerce && !arg.nil?
+    if self.respond_to?(:properties)
+      # in Chef 12.4+, properties replace attributes and these respond to
+      # coerce and default with a lazy block
+      def self.attribute(attr_name, validation_opts={})
+        lazy_default = validation_opts.delete(:lazy_default)
+        if lazy_default
+          Chef.log_deprecation("#{name} defines a lazy_default validation, instead specify default with a lazy block when Chef supports properties")
```
jkeiser: Rightest thing to do here is probably to change the callers to use `default: lazy { ... }`. You can declare that with this in SuperLWRP:

```ruby
def self.lazy(&block)
  Chef::DelayedEvaluator.new(&block)
end
```

And you can check for it in the `else` case like this:

```ruby
if validation_opts[:default].is_a?(Chef::DelayedEvaluator)
  lazy_default = validation_opts.delete(:default)
end
```

Then just change the callers to use it.

lib/chef/provisioning/aws_driver/aws_resource.rb
```diff
@@ -136,7 +136,14 @@ def self.get_aws_object_id(value, **options)
 
   protected
 
-  NOT_PASSED = Object.new
+  # This supports the introduction of the NOT_PASSED chef constant
```
jkeiser: This will work, but recall that in cheffish we fixed this with this instead:

1. Open up the `Chef` class directly so you get `NOT_PASSED` (instead of doing `module Chef::...` at the top, do `class Chef` and then `module ...`)

2. Define it like this:

```ruby
NOT_PASSED = Object.new if !defined?(NOT_PASSED)
```

lib/chef/provisioning/aws_driver/super_lwrp.rb
jkeiser: (This will allow you to not override `attribute` at all.)

tyler-ball: I think I'm trying to do the opposite of that - I don't want to convert `default: lazy { ... }` to `lazy_default`, I want to do the opposite.  I thought thats what I was doing with my current code.

I want to get rid of `lazy_default` in chef-provisioning-aws 2.0 so that we just use the common chef pattern.  What am I missing?

lib/chef/provisioning/aws_driver/aws_resource.rb
tyler-ball: :+1:

lib/chef/provisioning/aws_driver/aws_resource.rb
```diff
@@ -5,199 +5,208 @@
...
-
-  def initialize(name, run_context=nil)
-    name = name.public_send(self.class.aws_sdk_class_id) if name.is_a?(self.class.aws_sdk_class)
-    super
-    if run_context
-      driver run_context.chef_provisioning.current_driver
-      chef_server run_context.cheffish.current_chef_server
-    end
-  end
+class Chef
```
randomcamel: Github didn't do anything intelligent with this file's diff: is it just indenting?

tyler-ball: Set the [`?w=1`](https://github.com/chef/chef-provisioning-aws/pull/332/files?w=1) and the diff looks much better

randomcamel:
Looks reasonable, though I'm not in touch with the 12.5 changes, so that needs other eyes.

lib/chef/provisioning/aws_driver/super_lwrp.rb
jkeiser: In Provisioning, this is the syntax we use for lazy defaults:

```
attribute :x, lazy_default: proc { ... }
```

In 12.5 this is the syntax you use:

```
attribute :x, default: lazy { ... }
```

In the 12.5 side of that `if` statement, you are adding support to convert `lazy_default` to `default` inside `SuperLWRP`. I'm suggesting that instead, you remove `def self.attribute` entirely in 12.5 and convert all callers in `chef-provisioning-aws` to 12.5 syntax (`default: lazy { ... }`). This would just mean that the 12.4 side would have to stop directly supporting `lazy_default` as an option and instead recognize `default: lazy { ... }` and support that directly.

spec/aws_support/matchers/create_an_aws_object.rb
```diff
@@ -41,7 +41,7 @@ def match_failure_messages(recipe)
         differences += match_values_failure_messages(example.be_updated, recipe, "recipe")
 
         # Check for object existence and properties
-        resource = resource_class.new(name, nil)
+        resource = resource_class.new(name, recipe.client.run_context)
```
jkeiser: How necessary is this? I worry a tiny bit at giving the resource the ability to muck with the existing run_context. Not hugely, but some.

lib/chef/provisioning/aws_driver/super_lwrp.rb
tyler-ball: I guess I was trying to provide backwards compatibility for resources desiring to use `lazy_default`.  But you're right, I don't think _any_ resources leveraging `lazy_default` besides ones we have control of exist.

I'll check to make sure that `default: lazy { ... }` is backwards compatible with chef 12.4

spec/aws_support/matchers/create_an_aws_object.rb
tyler-ball: It is necessary so that we can [use the provider to look up the tags](https://github.com/chef/chef-provisioning-aws/blob/tball/chef_master/spec/aws_support/matchers/have_aws_object_tags.rb#L47-L53).

I'll remove it from the non-tag matchers

PR #334, opened 2015-09-21
V1 SDK automatically accepted ENV[AWS_CONFIG_FILE] but V2 doesn't, so we need to update to support that

tyler-ball:
\cc @jkeiser @cwebber @randomcamel @jaym

If `ENV["AWS_CONFIG_FILE"]` is nil we fallback to the default value (in `~/.aws`)


cwebberOps:
:+1:

randomcamel:
:+1: :shipit:

PR #337, opened 2015-09-22
bootstrap_options[:user_data] has a default on windows

hh:
Should fix #330 and and updates #325  #327

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -701,6 +701,13 @@ EOD
         bootstrap_options[:user_data] = Base64.encode64(bootstrap_options[:user_data])
       end
 
+      if machine_options[:is_windows]
+        Chef::Log.debug "Setting Default WinRM userdata for windows..."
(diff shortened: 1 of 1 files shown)


hh:
bootstrap_options[:user_data] is blank by default, so if we don't check for windows, we don't pull in the default windows user_data

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -701,6 +701,13 @@ def bootstrap_options_for(action_handler, machine_spec, machine_options)
         bootstrap_options[:user_data] = Base64.encode64(bootstrap_options[:user_data])
       end
 
+      if machine_options[:is_windows]
+        Chef::Log.debug "Setting Default WinRM userdata for windows..."
+        bootstrap_options[:user_data] = Base64.encode64(user_data) if bootstrap_options[:user_data].nil?
```
tyler-ball: Will you rename `user_data` to `default_windows_user_data` please?

tyler-ball:
This will need a spec to validate that a windows instance starting up without a `user_data` bootstrap_options gets a defaulted value

tyler-ball:
This was fixed in https://github.com/chef/chef-provisioning-aws/pull/400

PR #342, opened 2015-09-22
AwsRoute53HostedZone resource

randomcamel:
This is ready for 90% review. There are many, many additional widgets to add, but to limit PR size and make this available to folks who want it, this is a good basic pass.

TODO:
- [x] check for any need to poll on the change object returns from AWS calls. _[edit: not needed.]_
- [x] **update** doesn't catch if a RecordSet's `type` is changing.
- [x] **update** only checks for RecordSet presence/absence, not changes in fields.


chef-provisioning-aws.gemspec
```diff
@@ -24,6 +25,8 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec', '~> 3.0'
   s.add_development_dependency 'rake'
   s.add_development_dependency 'pry'
+  s.add_development_dependency 'pry-byebug'
```
tyler-ball: I think this is a great addition and we should do this everywhere

docs/examples/route53.rb
```diff
@@ -0,0 +1,73 @@
...
+# this will fail. we use the domain name as a data bag key, but Route 53 will add a trailing dot, and
+# furthermore Route 53 is content to have two HostedZones named "feegle.com.". in order to prevent unexpected
+# results, we prevent domain names from ending with a dot.
+aws_route53_hosted_zone "feegle.com."
+
+# create a Route 53 Hosted Zone (which AWS will normalize to "feegle.com.").
+aws_route53_hosted_zone "feegle.com"
+
+# create a Route 53 Hosted Zone with a CNAME record.
+# TODO(9/17/2015): maybe add an RRS attribute to append the enclosing zone name (like 'append_zone_name true').
```
tyler-ball: I think this attribute makes sense on the `aws_route53_hosted_zone` resource

lib/chef/provisioning/aws_driver/aws_resource.rb
```diff
@@ -118,6 +122,12 @@ def self.get_aws_object(value, resource: nil, run_context: nil, driver: nil, man
       resource = value
     else
       resource = new(value, run_context)
+
+      # populate AWS ID for resources like AwsRoute53HostedZone that don't have lazy-auto assignment.
+      if resource.public_send(resource.class.aws_id_attribute).nil?
```
tyler-ball: I don't think this is necessary - we overwrote `self.attribute` to set the `aws_id_attribute` to whatever attribute name is marked.

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -617,6 +623,10 @@ def rds
       @rds ||= AWS::RDS.new(config: aws_config)
     end
 
+    def route53
+      @route53 ||= AWS::Route53.new(config: aws_config)
```
tyler-ball: Lets just not add the V1 client - V2 or bust!

docs/examples/route53.rb
randomcamel: I'll have to run some experiments, but I think AWS assumes it.

randomcamel:
Nice to have, not sure it's needed for merging the first pass: It would be great to get `be_idempotent` working. Idempotence is easy to manually verify, so it's a matter of the code and Cheffish coming to terms.

lib/chef/resource/aws_route53_hosted_zone.rb
```diff
@@ -0,0 +1,220 @@
...
+  end
+
+  # since this is used exactly once, it could plausibly be inlined in #aws_object.
+  def get_record_sets_from_aws(hosted_zone_id, opts={})
+    params = { hosted_zone_id: hosted_zone_id }.merge(opts)
+    driver.route53_client.list_resource_record_sets(params)
+  end
+end
+
+class Chef::Provider::AwsRoute53HostedZone < Chef::Provisioning::AWSDriver::AWSProvider
```
tyler-ball: Is this the new way forward in the Chef 12.5 world?

lib/chef/resource/aws_route53_hosted_zone.rb
```diff
@@ -0,0 +1,220 @@
...
+  use_inline_resources
+
+  CREATE = "CREATE"
+  UPDATE = UPSERT = "UPSERT"
+  DELETE = "DELETE"
+  RRS_COMMENT = "Managed by chef-provisioning-aws"
+
+  attr_accessor :record_set_list
+
+  def make_hosted_zone_config(new_resource)
```
tyler-ball: The provider has access to the `new_resource` method, so this parameter is unecessary.  Was there a reason you were passing it in?

lib/chef/resource/aws_route53_hosted_zone.rb
```diff
@@ -0,0 +1,220 @@
...
+        if record_set_resources
+          change_list = record_set_resources.map { |rs| rs.to_aws_change_struct(CREATE) }
+
+          new_resource.driver.route53_client.change_resource_record_sets(hosted_zone_id: new_resource.aws_route53_zone_id,
+                                                                         change_batch: {
+                                                                           comment: RRS_COMMENT,
+                                                                           changes: change_list,
+                                                                           })
+        end
+      rescue => ex
```
tyler-ball: What do you think about catching `::Aws::Route53::Errors::ServiceError` and only deleting the zone then?  If there is a random Ruby error we don't necessarily need to delete the zone.

tyler-ball: I think we should also be logging the exception at debug level (in case there is another exception trying to delete the zone) and then log a message at info level saying there was an error creating the record sets and we're going to try to automatically delete the zone.

tyler-ball: Furthermore, this seems like a pretty good pattern when there are dependent, enclosed objects.  The convergent model says that we when the CCR is finished the aws object will match the resource or we will raise an error.  Because we couldn't match the resource, it seems really nice to try and clean up after ourselves instead of staying in a semi-broken state.  Although that isn't the worst because the user can just CCR again to converge.  Anyways, nice catch!

lib/chef/resource/aws_route53_hosted_zone.rb
```diff
@@ -0,0 +1,220 @@
...
+          # we already checked for duplicate Chef RR resources in #get_record_sets_from_resource.
+          keyed_chef_resources = record_set_resources.reduce({}) { |coll, rs| (coll[rs.aws_key] ||= []) << rs; coll }
+          keyed_aws_objects    = aws_record_sets.reduce({})      { |coll, rs| coll[rs.aws_key] = rs; coll }
+
+          keyed_chef_resources.each do |key, chef_resource_ary|
+            chef_resource_ary.each do |chef_resource|
+              # RR already exists...
+              if keyed_aws_objects.has_key?(key)
+                # ... but we want to delete or update it.
+                action = chef_resource.action.first == :destroy ? DELETE : UPDATE
```
tyler-ball: Will this always perform an update even if the record set already matches the resource?  I don't see the logic where we diff the current state to the desired state to see if we need an update.

lib/chef/resource/aws_route53_hosted_zone.rb
```diff
@@ -0,0 +1,220 @@
...
+      if rr_changes.size > 0
+        aws_struct = {
+          hosted_zone_id: hosted_zone.id,
+          change_batch: {
+            comment: "Purging RRs prior to deleting resource",
+            changes: rr_changes,
+          }
+        }
+
+        new_resource.driver.route53_client.change_resource_record_sets(aws_struct)
```
tyler-ball: Is this instantaneous, or do we need to poll and wait for it to occur?

lib/chef/resource/aws_route53_record_set.rb
```diff
@@ -0,0 +1,106 @@
+class Aws::Route53::Types::ResourceRecordSet
```
tyler-ball: This probably needs the leading `::`

lib/chef/resource/aws_route53_record_set.rb
```diff
@@ -0,0 +1,106 @@
...
+  # if you add the trailing dot, AWS returns "FATAL problem: DomainLabelEmpty encountered," so we'll stop that
+  # ourselves.
+  attribute :rr_name, required: true, callbacks: { "cannot end with a dot" => lambda { |n| n !~ /\.$/ }}
+  attribute :type, equal_to: %w(SOA A TXT NS CNAME MX PTR SRV SPF AAAA), required: true
+  attribute :ttl, kind_of: Fixnum, required: true
+
+  attribute :resource_records, kind_of: Array, required: true, is: lambda { |rr_list| validate_rr_type(action, rr_list, type) }
+
+  def initialize(name, *args)
+    self.rr_name(name) unless @rr_name
```
tyler-ball: What about using `default: lazy { name }` instead?

lib/chef/resource/aws_route53_record_set.rb
```diff
@@ -0,0 +1,106 @@
...
+      end
+    end
+
+    # TODO: be helpful and print out all duplicates, not just the first.
+
+    true
+  end
+end
+
+class Chef::Provider::AwsRoute53RecordSet < Chef::Provider::LWRPBase
```
tyler-ball: Is this provider even required since we pull the resources out of the `resource_collection`?

spec/integration/aws_route53_hosted_zone_spec.rb
```diff
@@ -0,0 +1,375 @@
+require 'spec_helper'
+
+describe Chef::Resource::AwsRoute53HostedZone do
+  extend AWSSupport
+
+  when_the_chef_12_server "exists", organization: 'foo', server_scope: :context do
+    with_aws "when connected to AWS" do
+
+      # even though these resources are demonstrably convergent, `be_idempotent` fails, including when
+      # `update_aws_object` is nothing but an empty `converge_by` block. suggestions welcome.
```
tyler-ball: I'll jump in and help troubleshoot this - we should figure it out today

spec/unit/chef/provisioning/aws_driver/route53_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'chef/provisioning/aws_driver/driver'
+
+describe Aws::Route53::Types::ResourceRecordSet do
+  it "returns the correct RecordSet unique key"
```
tyler-ball: I'm guessing these are still in progress?

lib/chef/resource/aws_route53_hosted_zone.rb
randomcamel: Which part?

lib/chef/resource/aws_route53_hosted_zone.rb
randomcamel: I thought I'd make it easier to unit-test (though of course I didn't do that in my unit-testing frenzy yesterday).

lib/chef/resource/aws_route53_hosted_zone.rb
randomcamel: Agreed about the logging. If there's any exception, though, even a Ruby error, as the code stands we can't guarantee that the resource has successfully converged... I could do this:

```ruby
      converged = false
      begin
        record_set_resources = get_record_sets_from_resource(new_resource)

        if record_set_resources
          change_list = record_set_resources.map { |rs| rs.to_aws_change_struct(CREATE) }
          new_resource.driver.route53_client.change_resource_record_sets ...

          converged = true
        end
      rescue ::Aws::Route53::Errors::ServiceError => ex
        # the change call is transactional, so we just need to clean up the zone we created.
        new_resource.driver.route53_client.delete_hosted_zone(id: zone.id)
        Chef::Log.error(ex)
        raise
      rescue => ex
        raise if !converged
      end
```

Or, alternately:

```ruby
      rescue => ex
        if !converged
          new_resource.driver.route53_client.delete_hosted_zone(id: zone.id)
        end
        Chef::Log.error(ex)
        raise
      end
```

What do you think?

lib/chef/resource/aws_route53_hosted_zone.rb
randomcamel: No, you're right, the test-and-repair is limited to whether a RecordSet is present or absent. I decided that in this iteration, the benefit of potentially saving one AWS call per HostedZone didn't outweigh the additional complexity. I'll add a comment to that effect, and start compiling a TODO list.

lib/chef/resource/aws_route53_record_set.rb
randomcamel: So far as I can tell, yes: it needs the `provides`, and then the compile phase evidently looks for an `action_create` method, so it needs the `action` blocks.

spec/integration/aws_route53_hosted_zone_spec.rb
randomcamel: That'd be rad. Let me implement polling first (though I could use pointers from you on that as well), since some bugs are being hidden by Route 53's willingness to have multiple zones with the same name.

spec/unit/chef/provisioning/aws_driver/route53_spec.rb
randomcamel: Yep.

lib/chef/resource/aws_route53_hosted_zone.rb
tyler-ball: Putting the provider in the same file as the resource

lib/chef/resource/aws_route53_hosted_zone.rb
tyler-ball: For unit testing reasons I am :+1:

lib/chef/resource/aws_route53_hosted_zone.rb
tyler-ball: Its not a _requirement_ that resources clan up after themselves if an error occurs.  Its only a requirement that they are convergent, so when the error is resolved they will move from whatever state they were left in to the desired state.

My fear is mostly rooted in the fact that we're proposing a new pattern here - if _anything_ goes wrong with convergence, try to rollback the convergence to the last known good state automatically.  But if there is some epidemic problem (like a ruby segfault, or we run OOM, or the disc drops out, etc.) we will try to perform a rollback and that, too, will probably fail.  So AWS is still in a non-desired state and there are a ton of errors in the log, making it difficult to see what happened.

It seems to me that there is a much better chance of recovering from the error if its only an AWS error.

Also, I have seen too many times that wildcard exception catching cause issues, so I'm naturally wary of it.

randomcamel: Hmm... since there's not an inoperable state we can be left in, and any resolution is to re-converge, then the right behavior is probably to just not catch anything, and let the exceptions fly.

lib/chef/resource/aws_route53_hosted_zone.rb
randomcamel: That happened when initially pairing with Mr. Keiser way back when, this is definitely part of his desired erasure of the resource/provider split. I don't know that I've seen a roadmap for that larger direction, though.

lib/chef/resource/aws_route53_hosted_zone.rb
randomcamel: The documentation says the status refers to whether the changes have propagated to all of AWS's DNS servers, but the AWS object creation is instantaneous, so no polling needed.

lib/chef/resource/aws_route53_hosted_zone.rb
tyler-ball: :+1:

docs/examples/route53.rb
```diff
@@ -0,0 +1,78 @@
...
+      type "CNAME"
+      ttl 3600
+      resource_records ["some-other-host"]
+    end
+  }
+end
+
+# Route 53 ResourceRecordSets are mostly analogous to DNS resource records (RRs). in the AWS Console they look
+# like first-class objects, but they don't have an AWS ID: only the RR name. (In other words, you can't have
+# an A record and a TXT record both pointing to 'snarf.feegle.com'.)
```
tyler-ball: snarf snarf

PR #344, opened 2015-09-25
some specs for elastic ips

redterror:
This started out with me wanting to add network interface support for ElasticIps, with the ultimate goal of using the ENI as part of the `bootstrap_options` for a new box (first boot w/ an eip!).

Then I realized I could `:ready` a box, attach the eip, and *then* converge it.  Good enough for me!


randomcamel:
:+1:

PR #349, opened 2015-09-29
support for configuring winrm https ala azure driver fixing #348

### LINKED ISSUE #348: support winrm over https transport

hh:
Fixes #348

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -812,10 +812,8 @@ EOD
 
     def create_winrm_transport(machine_spec, machine_options, instance)
       remote_host = determine_remote_host(machine_spec, instance)
+      winrm_transport_options = machine_options[:winrm_transport]
 
-      port = machine_spec.reference['winrm_port'] || 5985
-      endpoint = "http://#{remote_host}:#{port}/wsman"
-      type = :plaintext
       pem_bytes = get_private_key(instance.key_name)
       encrypted_admin_password = wait_for_admin_password(machine_spec)
 
@@ -823,14 +821,36 @@ EOD
       private_key = OpenSSL::PKey::RSA.new(pem_bytes)
       decrypted_password = private_key.private_decrypt decoded
 
-      winrm_options = {
+      shared_winrm_options = {
         :user => machine_spec.reference['winrm_username'] || 'Administrator',
-        :pass => decrypted_password,
-        :disable_sspi => true,
-        :basic_auth_only => true
+        :pass => machine_options[:password] || decrypted_password
       }
 
-      Chef::Provisioning::Transport::WinRM.new("#{endpoint}", type, winrm_options, {})
+      if(winrm_transport_options['https'] || winrm_transport_options[:https]  )
+        port = machine_spec.reference['winrm_port'] || '5986'
+        endpoint = "https://#{remote_host}:#{port}/wsman"
+        type = :ssl
+        winrm_options = {
+          :disable_sspi => winrm_transport_options['https'][:disable_sspi] || false, # default to Negotiate
+          :basic_auth_only => winrm_transport_options['https'][:basic_auth_only] || false, # disallow Basic auth by default
+          :no_ssl_peer_verification => winrm_transport_options['https'][:no_ssl_peer_verification] || false #disallow MITM potential by default
+        }
+      end
+
+      if(winrm_transport_options['http'] || winrm_transport_options[:http]  )
+        port = machine_spec.reference['winrm_port'] || '5985'
+        port = tcp_endpoint[:public_port] || '5986'
+        endpoint = "http://#{remote_host}:#{port}/wsman"
+        type = :plaintext
+        winrm_options = {
+          :disable_sspi => winrm_transport_options['http']['disable_sspi'] || false, # default to Negotiate
+          :basic_auth_only => winrm_transport_options['http']['basic_auth_only'] || false # disallow Basic auth by default
+        }
+      end
+
+      merged_winrm_options = winrm_options.merge(shared_winrm_options)
+      Chef::Provisioning::Transport::WinRM.new("#{endpoint}", type, merged_winrm_options, {})
+
     end
 
     def wait_for_admin_password(machine_spec)


lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -823,14 +821,36 @@ def create_winrm_transport(machine_spec, machine_options, instance)
...
+      end
+
+      if(winrm_transport_options['http'] || winrm_transport_options[:http]  )
+        port = machine_spec.reference['winrm_port'] || '5985'
+        port = tcp_endpoint[:public_port] || '5986'
+        endpoint = "http://#{remote_host}:#{port}/wsman"
+        type = :plaintext
+        winrm_options = {
+          :disable_sspi => winrm_transport_options['http']['disable_sspi'] || false, # default to Negotiate
+          :basic_auth_only => winrm_transport_options['http']['basic_auth_only'] || false # disallow Basic auth by default
```
randomcamel: These two are strings here, but symbols above?

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -812,10 +812,8 @@ def default_ami_for_region(region, criteria = {})
 
     def create_winrm_transport(machine_spec, machine_options, instance)
       remote_host = determine_remote_host(machine_spec, instance)
+      winrm_transport_options = machine_options[:winrm_transport]
```
tyler-ball: This is a new machine_option, correct?

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -823,14 +821,36 @@ def create_winrm_transport(machine_spec, machine_options, instance)
...
+        winrm_options = {
+          :disable_sspi => winrm_transport_options['https'][:disable_sspi] || false, # default to Negotiate
+          :basic_auth_only => winrm_transport_options['https'][:basic_auth_only] || false, # disallow Basic auth by default
+          :no_ssl_peer_verification => winrm_transport_options['https'][:no_ssl_peer_verification] || false #disallow MITM potential by default
+        }
+      end
+
+      if(winrm_transport_options['http'] || winrm_transport_options[:http]  )
+        port = machine_spec.reference['winrm_port'] || '5985'
+        port = tcp_endpoint[:public_port] || '5986'
```
randomcamel: Doesn't this overwrite `port`?

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -823,14 +821,36 @@ def create_winrm_transport(machine_spec, machine_options, instance)
...
+        endpoint = "http://#{remote_host}:#{port}/wsman"
+        type = :plaintext
+        winrm_options = {
+          :disable_sspi => winrm_transport_options['http']['disable_sspi'] || false, # default to Negotiate
+          :basic_auth_only => winrm_transport_options['http']['basic_auth_only'] || false # disallow Basic auth by default
+        }
+      end
```
randomcamel: This is tricky to DRY up, but if not that, some comments about the differences (some big, some small) would be helpful.

tyler-ball:
I'm super confused by this - where does `machine_spec.reference['winrm_port']` and `machine_spec.reference['winrm_username']` come from?  I don't see any references of them being set.

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -823,14 +821,36 @@ def create_winrm_transport(machine_spec, machine_options, instance)
...
+      shared_winrm_options = {
         :user => machine_spec.reference['winrm_username'] || 'Administrator',
-        :pass => decrypted_password,
-        :disable_sspi => true,
-        :basic_auth_only => true
+        :pass => machine_options[:password] || decrypted_password
       }
 
-      Chef::Provisioning::Transport::WinRM.new("#{endpoint}", type, winrm_options, {})
+      if(winrm_transport_options['https'] || winrm_transport_options[:https]  )
```
tyler-ball: If `winrm_transport_options` comes from `machine_options[:winrm_transport]` then I think we should just use symbols everywhere instead of the mix of strings and symbols.

The reason we use strings in `machine_spec.reference` is that it is converting them to/from a data_bag.  Otherwise, symbols are prefered.

hh: I was unsure on symbols vs strings, most of this is fork lifted from https://github.com/chef/chef-provisioning-azure/blob/master/lib/chef/provisioning/azure_driver/driver.rb#L253 

I'll update for symbols.

hh:
The bootstrap_options setting was brought over from chef-provisioning-azure:
I just forgot to change things over to symbols because I wasn't sure how everything is consumed:

https://github.com/chef/chef-provisioning-azure/blob/master/README.md#windows

```
with_machine_options bootstrap_options: {
                     :winrm_transport => { #optional
                       'https' => { #required (valid values: 'http', 'https')
                         :disable_sspi => false, #optional, (default: false)
                         :basic_auth_only => false, #optional, (default: false)
                         :no_ssl_peer_verification => true #optional, (default: false)
                       }
                     }
```

lib/chef/provisioning/aws_driver/driver.rb
hh: that double port thing was a leftover from the copypasta from chef-provisioning-azure, yeah it can be dried up a bit

lib/chef/provisioning/aws_driver/driver.rb
hh: no, typo / leftovers

lib/chef/provisioning/aws_driver/driver.rb
hh: That was brought over from https://github.com/chef/chef-provisioning-azure/blob/master/README.md#windows we probably need to decide on a cross provider way to specify all the azure stuff.... I don't think it goes into bootstrap, due to the fact that the way you connect might change later (after the first convergence, you might add ssh for example).

lib/chef/provisioning/aws_driver/driver.rb
hh: I was probably thinking about bringing symbols over, but got in a hurry. Symbols everywhere if they work is fine by me.

hh:
Closing due to new pull at #354

PR #352, opened 2015-10-01
Polishing touches for Route 53

randomcamel:
- [x] RecordSet names inherit the zone name automatically.
- [x] Variables from the declared scope are accessible inside RecordSets.
- [x] Parent zone has a `defaults` hash of values inherited by RecordSets.

### DIFF
--- a/lib/chef/resource/aws_route53_record_set.rb
+++ b/lib/chef/resource/aws_route53_record_set.rb
@@ -1,3 +1,20 @@
+#
+# Copyright:: Copyright (c) 2015 Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 class Aws::Route53::Types::ResourceRecordSet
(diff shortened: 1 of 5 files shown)


randomcamel:
@tyler-ball @jkeiser Please have a look when you get a chance.

lib/chef/resource/aws_route53_record_set.rb
```diff
@@ -53,25 +80,43 @@ def validate_rr_type(type, rr_list)
...
       true
     else
-      raise ArgumentError, "Argument '#{type}' must be one of #{%w(SOA A TXT NS CNAME MX PTR SPF AAAA)}"
+      raise ArgumentError, "Argument '#{type}' must be one of #{%w(A MX SRV CNAME TXT PTR AAAA SPF)}"
     end
   end
 
+  def validate_zone_name!(rr_name, zone_name)
+    if rr_name.end_with?('.') && rr_name !~ /#{zone_name}\.$/
+      raise(::Chef::Exceptions::ValidationFailed, "RecordSet name #{rr_name} does not match parent HostedZone name #{zone_name}.")
```
tyler-ball: If we didn't raise an error here, would AWS?  If so, was their error confusing?

spec/integration/aws_route53_hosted_zone_spec.rb
```diff
@@ -130,6 +130,36 @@
...
+                    ttl 3600
+                    resource_records ["some-other-host"]
+                  end
+                }
+              end
+            }.to create_an_aws_route53_hosted_zone("feegle.com",
+                                                   resource_record_sets: [{}, {}, sdk_cname_rr]).and be_idempotent
+          end
+
+          # AWS's error for this is "FATAL problem: DomainLabelEmpty encountered", so we help the user out.
```
tyler-ball: Nice improvement on a bad error, and thanks for the test that answered a question I had

tyler-ball:
This LGTM! :+1:

Can you also add these to the [Rakefile](https://github.com/chef/chef-provisioning-aws/blob/master/Rakefile#L42) so the tests are ran via CI?

PR #357, opened 2015-10-07
Fix deleting of AMIs, fixes #311

### LINKED ISSUE #311: machine_image recreation fails randomly
Simple code:

```ruby
machine_image "my-ami" do
  recipe "my-recipe"
  action [:destroy, :create]
end
```

Works in 50% of cases. In other 50 returns:

```
[2015-09-04T14:18:18+00:00] ERROR: machine_image[my-ami] (fyber-environments::_rebuild_base_images line 3) had an error: AWS::Core::Resource::NotFound: aws_image[base-ami] (ami-15745262) (basic_chef_client::block line 451) had an error: AWS::Core::Resource::NotFound: unable to find the instance
```

Is it related to the fact that AMIs always have `From-instance` tag that is never flushed and thus chef-provisioning-aws tries to destroy image from this tag even though it doesn't exist already?

Fodoj:
Previous implementation was checking against `AWS::EC2::Errors::InvalidInstanceID::NotFound` exception, which is listed as a valid exception in AWS docs. In reality, though, actual exception AWS returns is `AWS::Core::Resource::NotFound`. 

Previous implementation worked by accident, because if you create and then instantly delete an image, then instance is still around with terminated state, but if you try to create now and then delete in few hours, then  `AWS::Core::Resource::NotFound` will be returned and recipe will fail.

That the exact reason of this issue: https://github.com/chef/chef-provisioning-aws/issues/311 appearing. Now it's fixed.

### DIFF
--- a/spec/integration/machine_image_spec.rb
+++ b/spec/integration/machine_image_spec.rb
@@ -26,7 +26,7 @@ describe Chef::Resource::MachineImage do
         ).and be_idempotent
       end
 
-      describe 'action :destroy' do
+      describe 'action :destroy', :super_slow do
         with_converge {
           machine_image 'test_machine_image' do
             machine_options bootstrap_options: {
@@ -37,7 +37,7 @@ describe Chef::Resource::MachineImage do
           end
         }
 
-        it "deletes the image" do
+        it "destroys the image" do
           r = recipe {
             machine_image "test_machine_image" do
               action :destroy
@@ -47,9 +47,9 @@ describe Chef::Resource::MachineImage do
           ).and be_idempotent
         end
 
-        it "deletes the image if instance is gone long time ago" do
+        it "destroys the image if instance is gone long time ago" do
           image = driver.ec2_resource.images({filters: [ { name: "name", values: ["test_machine_image"] }]}).first
-          image.create_tags(tags: [{key: "from-instance", value: "i-4e5b71e8"}])
+          image.create_tags(tags: [{key: "from-instance", value: "i-12345678"}])
 
           r = recipe {
             machine_image "test_machine_image" do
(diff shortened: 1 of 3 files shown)


tyler-ball:
Great catch @Fodoj !

spec/integration/machine_image_spec.rb
```diff
@@ -25,6 +25,42 @@
           name: 'test_machine_image'
         ).and be_idempotent
       end
+
+      describe 'action :destroy' do
```
tyler-ball: This should be `describe 'action :destroy', :super_slow do` to mark this test as super slow

spec/integration/machine_image_spec.rb
```diff
@@ -25,6 +25,42 @@
...
+              action :destroy
+            end
+          }
+          expect(r).to destroy_an_aws_image('test_machine_image'
+          ).and be_idempotent
+        end
+
+        it "deletes the image if instance is gone long time ago" do
+          image = driver.ec2_resource.images({filters: [ { name: "name", values: ["test_machine_image"] }]}).first
+          image.create_tags(tags: [{key: "from-instance", value: "i-4e5b71e8"}])
```
tyler-ball: Will you make this `i-12345678` so it is more obvious that this should be a bogus id?

tyler-ball:
A few minor comments and one more request:  please add `machine_image_spec.rb` to the [Rakefile](https://github.com/chef/chef-provisioning-aws/blob/master/Rakefile#L42) so we start running this integration test through CI.

Don't worry about the Travis failures - its because Travis does not load secure environment variables for forks.  I'll run the CI when merging to master.

Fodoj:
@tyler-ball thanks for the review, changes are pushed

PR #365, opened 2015-10-13
Use Amazon API version 2 for RDS

josb:
This allows the use of "storage_encrypted: true" and other v2 API
options.

### DIFF
--- a/lib/chef/provider/aws_rds_instance.rb
+++ b/lib/chef/provider/aws_rds_instance.rb
@@ -26,7 +26,7 @@ class Chef::Provider::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSProvide
 
   def create_aws_object
     converge_by "create RDS instance #{new_resource.db_instance_identifier} in #{region}" do
-      new_resource.driver.rds.client.create_db_instance(options_hash)
+      new_resource.driver.rds_resource.create_db_instance(options_hash)
     end
   end
 
@@ -36,14 +36,12 @@ class Chef::Provider::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSProvide
     end
     # Wait up to 10 minutes for the db instance to shutdown
     converge_by "waited until RDS instance #{new_resource.name} was deleted" do
-      wait_for(
-        aws_object: instance,
-        query_method: :exists?,
-        expected_responses: [false],
-        acceptable_errors: [AWS::RDS::Errors::DBInstanceNotFound],
-        tries: 60,
-        sleep: 10
-      )
+      tries, sleep = 60, 10
+      Retryable.retryable(:tries => tries, :sleep => sleep, :not => [Aws::RDS::Errors::DBInstanceNotFound]) do |retries, exception|
+        instance.load
+        action_handler.report_progress "waited #{retries*sleep}/#{tries*sleep}s for #{instance.id} state to change to deleted (is: #{instance.db_instance_status})..."
+        raise StandardError if ['available', 'deleting'].include?(instance.db_instance_status)
+      end rescue Aws::RDS::Errors::DBInstanceNotFound # Swallow exception
     end
   end
 
--- a/lib/chef/provider/aws_rds_instance.rb
+++ b/lib/chef/provider/aws_rds_instance.rb
@@ -37,11 +37,11 @@ class Chef::Provider::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSProvide
     # Wait up to 10 minutes for the db instance to shutdown
     converge_by "waited until RDS instance #{new_resource.name} was deleted" do
       tries, sleep = 60, 10
-      Retryable.retryable(:tries => tries, :sleep => sleep, :not => [Aws::RDS::Errors::DBInstanceNotFound]) do |retries, exception|
+      Retryable.retryable(:tries => tries, :sleep => sleep, :not => [::Aws::RDS::Errors::DBInstanceNotFound]) do |retries, exception|
         instance.load
         action_handler.report_progress "waited #{retries*sleep}/#{tries*sleep}s for #{instance.id} state to change to deleted (is: #{instance.db_instance_status})..."
         raise StandardError if ['available', 'deleting'].include?(instance.db_instance_status)
-      end rescue Aws::RDS::Errors::DBInstanceNotFound # Swallow exception
+      end rescue ::Aws::RDS::Errors::DBInstanceNotFound # Swallow exception
     end
   end
 
--- a/lib/chef/resource/aws_rds_instance.rb
+++ b/lib/chef/resource/aws_rds_instance.rb
@@ -4,7 +4,7 @@ require 'chef/provisioning/aws_driver/aws_taggable'
 class Chef::Resource::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSRDSResource
   include Chef::Provisioning::AWSDriver::AWSTaggable
 
-  aws_sdk_type AWS::RDS::DBInstance, id: :db_instance_identifier
+  aws_sdk_type Aws::RDS::DBInstance, id: :db_instance_identifier
 
   attribute :db_instance_identifier, kind_of: String, name_attribute: true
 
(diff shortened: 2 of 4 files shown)


josb:
Not sure why those tests are complaining about credentials. It works fine here...

lib/chef/resource/aws_rds_instance.rb
```diff
@@ -28,14 +28,38 @@ class Chef::Resource::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSRDSReso
...
+    result = driver.rds_resource.db_instance(name)
+     begin
+      # try accessing it to find out if it exists
+      result.db_instance_status if result
+      result
+    rescue ::Aws::RDS::Errors::DBInstanceNotFound
+      result = nil
+    end
+    if result and result.db_instance_status != 'deleting'
+      result
```
Fodoj: small proposal to make this code a bit more compact:

```ruby
def aws_object
  result = self.driver.rds_resource.db_instance(name)
  return nil unless result && result.db_instance_status != 'deleting'
  result
  rescue ::Aws::RDS::Errors::DBInstanceNotFound
    nil
end
```

lib/chef/resource/aws_rds_instance.rb
```diff
@@ -28,14 +28,38 @@ class Chef::Resource::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSRDSReso
...
+    result = nil
+    if aws_object
+      begin
+        result = aws_object.db_instance_status
+      rescue ::Aws::RDS::Errors::DBInstanceNotFound
+        result = nil
+      end
+    end
+    result
+  end
```
Fodoj: another small proposal to make it a bit nicer and compact:

``` 
aws_object.db_instance_status if aws_object
rescue ::Aws::RDS::Errors::DBInstanceNotFound
  nil
end
```

tyler-ball:
@philbrookes We enabled integration tests on Travis but it uses encrypted Travis env variables, and these are only ran on branches in the original org (IE, `chef`).

I need to add something to the README

lib/chef/provider/aws_rds_instance.rb
```diff
@@ -36,14 +36,12 @@ def destroy_aws_object(instance)
...
-      wait_for(
-        aws_object: instance,
-        query_method: :exists?,
-        expected_responses: [false],
-        acceptable_errors: [AWS::RDS::Errors::DBInstanceNotFound],
-        tries: 60,
-        sleep: 10
-      )
+      tries, sleep = 60, 10
+      Retryable.retryable(:tries => tries, :sleep => sleep, :not => [Aws::RDS::Errors::DBInstanceNotFound]) do |retries, exception|
```
tyler-ball: All instances of `Aws::` should be `::Aws::` - otherwise we get a `NameError: Uninitialized Constant` error

lib/chef/provider/aws_rds_instance.rb
```diff
@@ -36,14 +36,12 @@ def destroy_aws_object(instance)
...
     converge_by "waited until RDS instance #{new_resource.name} was deleted" do
-      wait_for(
-        aws_object: instance,
-        query_method: :exists?,
-        expected_responses: [false],
-        acceptable_errors: [AWS::RDS::Errors::DBInstanceNotFound],
-        tries: 60,
-        sleep: 10
-      )
+      tries, sleep = 60, 10
```
tyler-ball: What was wrong with the `wait_for` logic?

lib/chef/resource/aws_rds_instance.rb
```diff
@@ -28,12 +28,21 @@ class Chef::Resource::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSRDSReso
...
 
   def aws_object
-    res = driver.rds.instances[name]
-    if res.exists? && ! ['deleted', 'deleting'].include?(res.status)
-      res
-    else
-      nil
-    end
+    result = self.driver.rds_resource.db_instance(name)
+    return nil unless result && result.db_instance_status != 'deleting'
```
tyler-ball: What if the status is `deleted`?  Does the V2 API return nil then?

lib/chef/resource/aws_rds_instance.rb
```diff
@@ -28,12 +28,21 @@ class Chef::Resource::AwsRdsInstance < Chef::Provisioning::AWSDriver::AWSRDSReso
...
+    result
+  rescue ::Aws::RDS::Errors::DBInstanceNotFound
+    nil
+  end
+
+  def endpoint
```
tyler-ball: I don't see this leveraged anywhere - what is the reason for adding it?

lib/chef/provider/aws_rds_instance.rb
josb: The exists? method no longer exists, and I couldn't find another way to detect whether the database had been deleted. The expected_response to the ```instance.load``` is the ```::Aws::RDS::Errors::DBInstanceNotFound``` exception but I didn't see a way to make ```wait_for``` use it the right way. If you have a better way, I'm all ears.

lib/chef/resource/aws_rds_instance.rb
josb: As far as I know, yes. I have not seen it return ```deleted```.

lib/chef/resource/aws_rds_instance.rb
josb: The original driver had it so I made it work. Isn't this present so the Recipe DSL can get at the backing AWS object?

josb:
So the encrypted Travis environment variables is why my build is failing? Because the patch has been working fine here for multiple people...

lib/chef/resource/aws_rds_instance.rb
josb: I should add that ```endpoint``` and ```db_instance_status``` were added because I needed them, notably for getting the endpoint value so that clients can connect to the RDS instance provisioned. I understand this may have some ramifications for resource cloning, but I'm not sure exactly what, so any insight will be appreciated.

josb:
What else is needed to make this land so I don't have to patch my ChefDK? Thanks.

lib/chef/resource/aws_rds_instance.rb
tyler-ball: The only method needed to get at the backing AWS object is the `aws_object` method on the Resource.  When I do a `ctrl+f` on this page, I do not use any reference to `endpoint` being called.  Thats why I asked whether this method was needed.  `db_instance_status` is used elsewhere so I understand having that method.

We disable resource cloning for chef-provisioning resources, so that isn't an issue.

josb: Fair enough. I'll replace ```endpoint``` in my code with ```aws_object.endpoint```. Thanks, Tyler!

lib/chef/provider/aws_rds_instance.rb
jkeiser: Can `wait_for` take a proc? Maybe something like `query_method: proc { ['available', 'deleting'].include?(instance.db_instance_status) }`?

lib/chef/resource/aws_rds_instance.rb
jkeiser: FYI: http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Status.html

Looks like no `deleted`.

lib/chef/provider/aws_rds_instance.rb
josb: ```wait_for``` does ```current_response = aws_object.send(query_method)``` so I don't think a proc will work. The code should basically wait for ```AWS::RDS::Errors::DBInstanceNotFound``` and I don't think ```wait_for``` can do that out of the box.

tyler-ball: I addressed this in https://github.com/chef/chef-provisioning-aws/blob/86fda7c7a882dc3fffc3508c5fb8b9d6a6248a26/lib/chef/provider/aws_rds_instance.rb#L39-L48 - I made `wait_for` take an optional block that is called before we execute the query.

I could totally see us modifying `wait_for` to require a block that does the _actual_ checking instead of using the `query_method` parameter.  Thats @jkeiser's suggestion and I agree with it.

josb: Thanks.

But doesn't ```acceptable_errors``` just swallow the exception and not exit the ```wait_for``` call?

tyler-ball: @josb https://github.com/nfedyashev/retryable#description

Nope - because we catch `acceptable_errors` and squelch them (don't reraise), the retryable block will exit.  Have you seen different behavior?  The tests for this were passing for me locally and not taking the full timeout (10 minutes).

josb: @tyler-ball I didn't see a description of ```acceptable_errors``` on that page, so I looked at the code and saw

    rescue *acceptable_errors
    end

which led me to believe that option would just swallow the exception and not finish (since the block to ```Retryable.retryable``` seems to need to raise an exception in order to exit?). That option could use some docs for sure on how it behaves.

If it does work, then that's great :)

PR #372, opened 2015-10-16
Fix index field creation during cloudsearch create.

tylercloke:


stevendanna:
:+1:

randomcamel:
:+1: (Travis break is unrelated.)

Issue #369, opened 2015-10-16
with_driver "aws::us-west-2" block breaking VPC launch in loop


djapg:
When attempting to launch multiple VPCs in a recipe I had working successfully a few months ago, I get:

```ruby
require 'chef/provisioning/aws_driver'
with_chef_server 'https://mychefserver.com/organizations/myorg',
  :client_name => Chef::Config[:node_name],
  :signing_key_filename => "#{ENV['HOME']}/.chef/#{ENV['CHEF_USER']}.pem"

regions = { 'us-west-2' => '192.168.0.0/24',
            'eu-west-1' => '192.168.1.0/24' }

regions.each do |region,cidr|
  with_driver "aws::#{region}" do
    vpc = aws_vpc region do
      cidr_block cidr
      enable_dns_support true
      enable_dns_hostnames true
      internet_gateway true
      main_routes '0.0.0.0/0' => :internet_gateway
    end
  end
end
```
I get 

```
NoMethodError
    -------------
    aws_internet_gateway[igw-managed-by-vpc-0d1a4d68] (no AWS object id) (basic_chef_client::block line 208) had an error: NoMethodError: undefined method `driver_url' for nil:NilClass
```
I can pull out the each, and get the same error:
```
region = 'us-west-2'
cidr = '192.168.0.0/24'

with_driver "aws::#{region}" do
  vpc = aws_vpc region do
    cidr_block cidr
    enable_dns_support true
    enable_dns_hostnames true
    internet_gateway true
    main_routes '0.0.0.0/0' => :internet_gateway
  end
end
```
until I remove the do/end and the single VPC builds fine:

```
region = 'us-west-2'
cidr = '192.168.0.0/24'

with_driver "aws::#{region}"

vpc = aws_vpc region do
  cidr_block cidr
  enable_dns_support true
  enable_dns_hostnames true
  internet_gateway true
  main_routes '0.0.0.0/0' => :internet_gateway
end
```
Unfortunately I can't use the with_driver on its own in the loop
```
regions.each do |region,cidr|

  with_driver "aws::#{region}"

  vpc = aws_vpc region do
    cidr_block cidr
    enable_dns_support true
    enable_dns_hostnames true
    internet_gateway true
    main_routes '0.0.0.0/0' => :internet_gateway
  end

end
```

This time I get a different error

```
[...]

PR #373, opened 2015-10-16
Passing drivers to the inline_resource

tyler-ball:
Fixes https://github.com/chef/chef-provisioning-aws/issues/369

The [`with_driver`](https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/chef_run_data.rb#L54) code clears the set driver once its block has executed.  Unfortunately that means the correct driver is only available during compile time - any resource created after compile time (like `Cheffish.inline_resource` does) loses the driver scope.


randomcamel:
:+1:

PR #385, opened 2015-10-30
Add aws_elasticsearch_domain resource

stevendanna:


sdelano:
LGTM, but I'm not qualified to sign off on this.

danielsdeleo:
LGTM, too, but I'm not familiar with the inner workings of chef-provisioning

lib/chef/resource/aws_elasticsearch_domain.rb
```diff
@@ -0,0 +1,44 @@
...
+  class Elasticsearch
+    class Domain
+    end
+  end
+end
+
+class Chef::Resource::AwsElasticsearchDomain < Chef::Provisioning::AWSDriver::AWSResource
+  aws_sdk_type ::AWS::Elasticsearch::Domain
+
+  INSTANCE_TYPES = %w(m3.medium.elasticsearch m3.large.elasticsearch
```
tyler-ball: If the user specifies an invalid instance_type, what does the error look like from AWS?  I prefer to let these kind of validations fall through to AWS (if their error is informative) so we don't have to keep our code in sync with their updated instance types.

spec/integration/aws_elasticsearch_domain_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'spec_helper'
+
+describe Chef::Resource::AwsElasticsearchDomain do
+  extend AWSSupport
+  when_the_chef_12_server "exists", organization: "foo", server_scope: :context do
+    with_aws "when connected to AWS" do
+
+      time = DateTime.now.strftime('%Q')
+
+      it "aws_elasticsearch_domain 'test-#{time}' creates a elasticsearch domain" do
```
tyler-ball: Thanks for writing a spec for this!  Can you add some more coverage to match our [resource acceptance criteria](https://github.com/chef/chef-provisioning-aws/blob/master/CONTRIBUTING.md#resourceprovider-acceptance-criteria)?  Basically, you need 1 test that sets all possible attributes, 1 test that checks that on an existing object all possible fields are updatable, and 1 test to validate the `:destroy` action.

An example of this is [load_balancer](https://github.com/chef/chef-provisioning-aws/blob/13adbadf798f562b8a9c76a0a60a5dd1383b47f6/spec/integration/load_balancer_spec.rb#L27-L269) specs (although yours should be simpler because your resource does not reference other resources).

lib/chef/resource/aws_elasticsearch_domain.rb
randomcamel: We should note in the docs that these are what we support, since presumably Amazon will change the list.

spec/integration/aws_elasticsearch_domain_spec.rb
stevendanna: D'oh thanks for the pointer, apologies for not reading that document before.

tyler-ball:
There is some additional test coverage needed, but the resource and provider LGTM!

tyler-ball:
Also, you should add the test file to the [list](https://github.com/chef/chef-provisioning-aws/blob/994d25daa9912daaf991f41966670b714775541a/Rakefile#L42) of tests executed by Travis.

lib/chef/resource/aws_elasticsearch_domain.rb
stevendanna: I'll check what the API returns.  I went with this initially because the instance types are not just EC2 instance types.  If the API returns some hint about what is se-table, I agree we should just let it fall through.

spec/integration/aws_elasticsearch_domain_spec.rb
tyler-ball: No worries, I don't think we did a good job publishing its existence

randomcamel:
:+1: with Tyler's comments.

lib/chef/provider/aws_elasticsearch_domain.rb
```diff
@@ -0,0 +1,131 @@
...
+
+  def converge_tags
+    aws_tagger.converge_tags
+  end
+
+  private
+
+  def required_updates(object)
+    ret = []
+    ret << "   update cluster configuriguration" if cluster_options_changed?(object)
```
marcparadise: `configuration`

stevendanna: gracias

stevendanna:
I've added some testing, added aws_tag support, fixed up the instance list, and added this file to the rakefile.  The test cases don't cover as much as I want, but the elasticsearch API limits appear to be very low so I've cut the test cases to make sure we can run them reliably

lib/chef/provider/aws_elasticsearch_domain.rb
```diff
@@ -0,0 +1,131 @@
...
+                    end
+  end
+
+  def converge_tags
+    aws_tagger.converge_tags
+  end
+
+  private
+
+  def required_updates(object)
```
jkeiser: Man, we really need to get us some compat_resource on up in here.

jkeiser:
@stevendanna I think we need to know some things about patchiness. Do you want `desired` to be patchy? And if so (I personally think it's a generally good idea, but discretion is required), is the Amazon SDK for update_elasticsearch_domain_config actually patchy?  Specifically, what happens if you do this sequence of actions: 

```ruby
# 1. Create the domain
aws_elasticsearch_domain 'clowny' do
end

# 2. Update everything in the domain
aws_elasticsearch_domain name do
  instance_type "m3.medium.elasticsearch"
  instance_count 2
  dedicated_master_enabled true
  dedicated_master_type "m3.medium.elasticsearch"
  dedicated_master_count 2
  zone_awareness_enabled true
  ebs_enabled true
  volume_type "io1"
  volume_size 35
  iops 1000
  automated_snapshot_start_hour 2
  access_policies policy(name)
  aws_tags key1: "value"
end

# 3. Update the domain with empty.
aws_elasticsearch_domain 'clowny' do
end

# 4. Update the domain just enough to trigger some changed? calls.
aws_elasticsearch_domain 'clowny' do
  instance_type "m3.medium.elasticsearch"
  volume_size 66
end
```

Patchy, for reference, means "if the user does not specify a property, its value does not change."

I expect the state after (3) and (4) to be exactly the same as after (2), if this is patchy; I expect the state after (3) to be the same as (1) (and for (4) to be (3) + just those two changes) if it is not patchy. I don't expect anything in between personally. If the AWS API is patchy, then I think your code is patchy. I just recall a lot of AWS APIs not being patchy, so I worry.

spec/integration/aws_elasticsearch_domain_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+  end
+end
+
+describe Chef::Resource::AwsElasticsearchDomain do
+  extend AWSSupport
+
+  let(:all_options_result) do
+    {created: true,
+     elasticsearch_cluster_config: {
+       instance_type: "m3.medium.elasticsearch",
```
jkeiser: Is this the cheapest one available? (Or is the default the cheapest, and this is the second cheapest so we can test that we can change it?)

spec/integration/aws_elasticsearch_domain_spec.rb
```diff
@@ -0,0 +1,119 @@
+require 'spec_helper'
+
+def policy(name)
+  <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Deny",
```
tyler-ball: Actually, I think the reason the user is failing to communicate is because the effect here is `Deny`.  Setting the access policy to allow all from the UI gives me a document that looks like

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": [
          "*"
        ]
      },
      "Action": [
        "es:*"
      ],
      "Resource": "arn:aws:es:us-east-1:#{user_id}:domain/#{name}/*"
    }
  ]
}
```

tyler-ball: You should be able to get `user_id` with `user_id = iam_client.get_user.user.user_id`

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -644,6 +644,10 @@ def cloudsearch(api_version="20130101")
       @cloudsearch[api_version]
     end
 
+    def elasticsearch
+      @elasticsearch ||= Aws::ElasticsearchService::Client.new(region: aws_config.region)
```
tyler-ball: This should follow the pattern from https://github.com/chef/chef-provisioning-aws/blob/994d25daa9912daaf991f41966670b714775541a/lib/chef/provisioning/aws_driver/driver.rb#L655 and just pass the entire `aws_config_2` hash to the client.

Actually, will you just change [`AWS_V2_SERVICES.each`](https://github.com/chef/chef-provisioning-aws/blob/994d25daa9912daaf991f41966670b714775541a/lib/chef/provisioning/aws_driver/driver.rb#L651) to look like

```ruby
    AWS_V2_SERVICES.each do |load_name, short_name|
      class_eval <<-META
      def #{short_name}_client
        @#{short_name}_client ||= ::Aws::#{load_name}::Client.new(**aws_config_2)
      end
      META
      
      if defined? ::Aws::#{load_name}::Resource
      class_eval <<-META
        def #{short_name}_resource
          @#{short_name}_resource ||= ::Aws::#{load_name}::Resource.new(**(aws_config_2.merge({client: #{short_name}_client})))
        end
      META
```

and add `"ElasticsearchService" => "elasticsearch"` to the [`AWS_V2_SERVICES`](https://github.com/chef/chef-provisioning-aws/blob/994d25daa9912daaf991f41966670b714775541a/lib/chef/provisioning/aws_driver/driver.rb#L31) hash then you don't need to define this here

spec/integration/aws_elasticsearch_domain_spec.rb
stevendanna: This is the cheapest and is also the default which is slightly cheating.

spec/integration/aws_elasticsearch_domain_spec.rb
stevendanna: @tyler-ball Yeah, I'm not sure on the permissions.  The ID you see in the resource is actually the account ID.  I've updated the policy to be an Allow policy for the specific principal based on the ARN of the configured user, but am still seeing the failure.  AWS access policies can take over 10 minutes to apply so my guess is that we have to ensure the test user has access by default.

stevendanna:
@tyler-ball I believe all of the elasticsearch tests are now passing with the updated policy.  The overall test run is failing, but that appears to be problems with the loadbalancer resource which this shouldn't have touched.

@jkeiser As you mention, whether or not this is patchy is largely a function of whether the AWS API is patchy since we only include an option if the user specified it.  The AWS API does appear to be patchy in this case:

```
irb(main):018:0> c.describe_elasticsearch_domain({domain_name: "ssd-cs-elasticache"}).domain_status.snapshot_options
=> #<struct Aws::ElasticsearchService::Types::SnapshotOptions automated_snapshot_start_hour=2>
irb(main):019:0> c.update_elasticsearch_domain_config({domain_name: "ssd-cs-elasticache", elasticsearch_cluster_config: { instance_count: 2}})
irb(main):020:0> c.describe_elasticsearch_domain({domain_name: "ssd-cs-elasticache"}).domain_status.snapshot_options
=> #<struct Aws::ElasticsearchService::Types::SnapshotOptions automated_snapshot_start_hour=2>
```

tyler-ball:
@stevendanna Thanks for the PR!

PR #387, opened 2015-11-06
ElasticsearchService requires aws-sdk 2.1.26+

stevendanna:


tyler-ball:
Imma' merge this but I don't think its the issue of the error you reported

PR #389, opened 2015-11-11
fix a bug where an existing zone with a comment would explode if no comment was given

redterror:
Title pretty much says it all.  I had an existing zone (created by AWS's registrar) and when I tried to manage it with chef-provisioning-aws, things broke.

Side question - are there plans to support ALIAS records for R53?

### DIFF
--- a/spec/integration/aws_route53_hosted_zone_spec.rb
+++ b/spec/integration/aws_route53_hosted_zone_spec.rb
@@ -64,6 +64,17 @@ describe Chef::Resource::AwsRoute53HostedZone do
             }.to create_an_aws_route53_hosted_zone(zone_name,
                                                    config: { comment: expected_comment }).and be_idempotent
           end
+
+          it "updates the zone comment when none is given" do
+            expect_recipe {
+              aws_route53_hosted_zone zone_name do
+                comment "Initial comment."
+              end
+              aws_route53_hosted_zone zone_name do
+              end
+            }.to create_an_aws_route53_hosted_zone(zone_name,
+                                                   config: { comment: nil }).and be_idempotent
+          end
         end
 
         context "RecordSets" do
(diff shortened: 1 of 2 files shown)


randomcamel:
We're still working out a roadmap for chef-provisioning-aws, so we don't have a plan for ALIAS records currently, but looking at [the API](http://docs.aws.amazon.com/sdkforruby/api/Aws/Route53/Client.html#change_resource_record_sets-instance_method), it may not be hard to knock out, given the rest of the structure is now in place.

spec/integration/aws_route53_hosted_zone_spec.rb
```diff
@@ -64,6 +64,17 @@
...
+
+          it "updates the zone comment when none is given" do
+            expect_recipe {
+              aws_route53_hosted_zone zone_name do
+                comment "Initial comment."
+              end
+              aws_route53_hosted_zone zone_name do
+              end
+            }.to create_an_aws_route53_hosted_zone(zone_name,
+                                                   config: { comment: nil }).and be_idempotent
```
randomcamel: What was the error your code was exploding with? I'm not sure how this test exercises the `default` value.

redterror: Here's a snippit:

```
[2015-11-11T01:21:26+00:00] INFO: [Aws::Route53::Client 400 0.028024 0 retries] update_hosted_zone_comment(id:"MY-ZONE-ID",comment:nil) Aws::Route53::Errors::MalformedInput Unexpected complex element termination



    ================================================================================
    Error executing action `create` on resource 'aws_route53_hosted_zone[EXAMPLE.COM] (/hostedzone/MY-ZONE-ID)'
    ================================================================================

    Aws::Route53::Errors::MalformedInput
    ------------------------------------
    Unexpected complex element termination
```

By not passing a `comment`, the default is used.  If you want to be explicit then it can be written as:

```
aws_route53_hosted_zone zone_name do
  comment nil
end
```
The awkward thing is that `nil` still comes back from the API.

PR #390, opened 2015-11-11
Fixes #367.

### LINKED ISSUE #367: Bug when changing existing routes on aws_route_table converge
Referencing #353, @lfrodrigues encountered a bug in v.1.5.1 when converging an aws_route_table that was replacing an existing route (see #353 for the full backtrace).

```
undefined method `target' for #<Aws::EC2::Route:0x0000000933c0f8>
```

```
NoMethodError: aws_route_table[route-table-management] (rtb-1d7d2a78) (cluster_phase1::machine_internetnat line 55) had an error: NoMethodError: undefined method `target' for #<Aws::EC2::Route:0x00000
00caf0210>
```

My analysis was:

> It seems like this line should reference current_target instead of current_route.target (AWS::EC2::Route doesn't have a target attribute or method, not sure if this existing previously). This seems to happen when the route you are setting differs from what is there, so a quick fix might be to remove it or change it in the AWS console first before converging.

joaogbcravo:
Replaces inexistent 'current_route.target' by current_target.


lskillen:
:+1: I was just about to submit the same pull request.

PR #391, opened 2015-11-11
Route table with vgw

joaogbcravo:


tyler-ball:
Thanks @joaogbcravo !  I accepted it without a test because we don't have a resource for `virtual_private_gateway` at this point, so users will have to specify strings

joaogbcravo:
Yup, thanks to accept it! :) When I have time, if no one do it first, I will do the vpw resource/provider.

Cheers

PR #397, opened 2015-11-16
Don't iterate through CloudSearch index_fields if none were given

stevendanna:


stevendanna:
Not really sure how master was passing as this was failing the currently present tests.

tyler-ball:
@stevendanna Because the old [Rakefile](https://github.com/chef/chef-provisioning-aws/blob/master/Rakefile#L40-L50) was only running a specified subset of tests.  Now I added _all_ the non `super_slow` tests to Travis so it should run them now.

stevendanna:
Ahh that makes sense

lib/chef/provider/aws_cloudsearch_domain.rb
```diff
@@ -127,8 +127,10 @@ def create_index_field(field)
   end
 
   def create_index_fields
-    new_resource.index_fields.each do |field|
-      create_index_field(field)
+    if ! new_resource.index_fields.nil?
```
tyler-ball: Can you make this `unless` instead of `if !` please?  Then I'll accept this PR - I've got the tests running locally and I'll work on fixing travis while merging some more of these PRs

stevendanna: Done

tyler-ball:
I'm going to go ahead and merge this and work on test failures in another PR  - Thanks @stevendanna !

PR #401, opened 2015-11-17
fix aws_rds_subnet_group example

keen99:
example was using wrong attribute name, corrected according to attribute defined in 
https://github.com/chef/chef-provisioning-aws/blob/master/lib/chef/resource/aws_rds_subnet_group.rb#L11


tyler-ball:
Thanks @keen99 !

PR #407, opened 2015-11-25
Support for RDS parameter groups.

tylercloke:
Idempotence is impossible atm due to a bug in the amazon API (see the relevant comment in the code).

It is being fixed by Amazon, but it will be awhile before it is out in the wild. Once it is fixed, I will come back and make this resource idempotent, but this works for now.


lib/chef/provider/aws_rds_parameter_group.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    ret.unshift("update RDS parameter group #{new_resource.name} in #{region}") unless ret.empty?
+    ret
+  end
+
+
+  private
+
+  def update_parameters(parameter_group)
```
stevendanna: Style nitpick: Most places we try to use a `?` at the end of the method name to denote predicate methods.

lib/chef/provider/aws_rds_parameter_group.rb
```diff
@@ -0,0 +1,133 @@
...
+    # they are both arrays, so to compare current to wanted parameter hashes, we must iterate in O(n^2) :'(
+    return false if ! (desired_options[:parameters].nil? || desired_options[:parameters].empty?)
```
stevendanna: Is this logic right?  If `desired_options[:parameters]` is `nil` shouldn't update_parameters return false?

stevendanna:
:+1: This looks reasonable to me.

Issue #419, opened 2015-12-17
Add ECS service and task definition resource type


cwndrws:
It would be really nice to be able to use this with EC2 Container Service.

thehar:
Bumping this.

PR #424, opened 2016-01-07
Add ECS to Resources not started yet

cwandrews:
This is to address #419


drywheat:
It baffles me that these resources have NOT yet been implemented. You run the risk of people migrating to Terraform in order to codify this aspect of their infrastructure.

Issue #420, opened 2015-12-17
Add Health Checks to loadbalancers


cwndrws:
It doesn't seem like there's a way to configure a health check. I would like to be able to configure a health check and then attach it to a load balancer, or just declare a health check on load balancer creation.

rgeddis:
Not sure if this exactly what you are looking but you can configure the health check in the load_balancer_options, for example

{code}                                                                                                                                                                                        load_balancer_options lazy {                                                                                                                                                                                        
  {                                                                                                                                                                                                                 
    listeners: [{                                                                                                                                                                                                   
      port: 80,                                                                                                                                                                                                     
      protocol: :http,                                                                                                                                                                                              
      instance_port: 80,                                                                                                                                                                                            
      instance_protocol: :http                                                                                                                                                                                      
    }, {                                                                                                                                                                                                            
      port: 443,                                                                                                                                                                                                    
      protocol: :https,                                                                                                                                                                                             
      instance_port: 443,                                                                                                                                                                                           
      instance_protocol: :https                                                                                                                                                                    
    }],                                                                                                                                                                                                             
    health_check: {                                                                                                                                                                                                 
      healthy_threshold: 2,                                                                                                                                                                                         
      unhealthy_threshold: 4,                                                                                                                                                                                       
      interval: 12,                                                                                                                                                                                                 
      timeout: 5,                                                                                                                                                                                                   
      target: 'HTTPS:443/_status'                                                                                                                                                                                   
    },                                                                                                                                                                                                              
    scheme: 'internet-facing',                                                                                                                                                                                      
    subnets: lb_subnet,                                                                                                                                                                                             
    security_groups: [                                                                                                                                                                                              
      security_group.aws_object.id                                                                                                                                                                                   
    ]                                                                                                                                                                                                               
  }                                                                                                                                                                                                                 
}                                                                                                                                                                                   
{code}

cwandrews:
@rgeddis where are you seeing this in the docs? I don't see it in the docs for this repo nor in the [aws docs](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/ELB/Client.html#create_load_balancer-instance_method) that describe the load balancer creation options.

lskillen:
[...]

PR #426, opened 2016-01-08
Add a health check to the load balancer options example

cwandrews:
This adds an example of adding a health check to a load balancer in the `load_balancer_options` in the example in the docs as discussed in #420


joaogbcravo:
Nice! Thanks 👍

Issue #427, opened 2016-01-08
VPC peering - cannot peer to outside VPCs


keen99:
Attempting to create a aws_vpc_peering_connection to a VPC that's not managed by chef-provisioning (and therefor doesn't have a data bag) results in failure.

```
had an error: RuntimeError: Chef::Resource::AwsVpc[centraltech] does not exist!

had an error: RuntimeError: Chef::Resource::AwsVpc[vpc-48123462d] does not exist!
```

These lookups happen here:

https://github.com/chef/chef-provisioning-aws/blob/13e69c444b95aac2f41ee7f0bd307f13bba33734/lib/chef/provider/aws_vpc_peering_connection.rb#L30

and here:

https://github.com/chef/chef-provisioning-aws/blob/13e69c444b95aac2f41ee7f0bd307f13bba33734/lib/chef/provider/aws_vpc_peering_connection.rb#L58-L59

Clearly the intent here is to allow the more friendly vpc "Name" tag to be used to define the peering relationship.   That's great....but the actual API just requires a vpc-x123x5 style ID.

PR incoming to support passing real vpc IDs both for new_resource.vpc and new_resource.peer_vpc - in those cases we'll avoid the lookup.

Not sure where to start (and running out of week...) for adding tests to test these cases though.

badmadrad:
I agree that passing the vpc id would be nice but I actually went for a different approach. I just created the data bag myself. Created a json file like so:

    {
        "id": "my_vpc_name",
        "reference": {
        "id": "vpc-M2343249"
        },
        "driver_url": "aws::us-east-1"
    }
Saved as **mydatabag.json**. Then:

`knife data bag from file aws_vpc mydatabag.json`

And I was able to call by my friendly vpc name.

    aws_vpc_peering_connection "peer1" do 
      vpc "chef_vpc"
      peer_vpc 'my_vpc_name'
    end

keen99:
> I just created the data bag myself. Created a json file like so:


Great workaround for those who need to do this but cant/wont run a patched
version, thanks!

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #416, opened 2015-12-15
Fix PeeringConnection:

joaogbcravo:
When we have a vpc that belongs to some other account, we can't lookup it.

PR #428, opened 2016-01-08
allow vpc peering to accept explicit VPC IDs instead of friendly names

keen99:
to avoid lookups against resources we might not maintain - #427

### DIFF
--- a/lib/chef/provider/aws_vpc_peering_connection.rb
+++ b/lib/chef/provider/aws_vpc_peering_connection.rb
@@ -27,7 +27,15 @@ class Chef::Provider::AwsVpcPeeringConnection < Chef::Provisioning::AWSDriver::A
     options[:vpc_id] = new_resource.vpc
     options[:peer_vpc_id] = new_resource.peer_vpc
     options[:peer_owner_id] = new_resource.peer_owner_id unless new_resource.peer_owner_id.nil?
-    options = AWSResource.lookup_options(options, resource: new_resource)
+
+    ## for vpc and peer_vpc, don't use a lookup if they're a string that is an actual VPC identifier - it's entirely possible we dont know about it in our scope (ie, 3rd party account or manage by separate tools)
+    unless new_resource.vpc.class == String && new_resource.vpc.to_s =~ /^vpc-.*/
+      options[:vpc_id] = Chef::Resource::AwsVpc.get_aws_object_id(new_resource.vpc, resource: new_resource)
+    end
+    unless new_resource.peer_vpc.class == String && new_resource.peer_vpc.to_s =~ /^vpc-.*/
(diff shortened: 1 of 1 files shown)


joaogbcravo:
Hey!
#416 fix this problem! 

Cheers,
João

keen99:
I saw yours after I'd already built mine - but you didn't cover the update case, only create.  ie update_aws_object - https://github.com/chef/chef-provisioning-aws/blob/13e69c444b95aac2f41ee7f0bd307f13bba33734/lib/chef/provider/aws_vpc_peering_connection.rb#L58-L59

See issue #427 for more detail.

keen99:
also, #428 goes a bit further by allowing both "vpc" and "peer_vpc" to be specific IDs and avoid lookups (since you might want to define a peering relationship on a VPC not created/maintained by chef)

joaogbcravo:
Hey,

Probably I'm not correctly seeing problem and the resolution, but #416 have the following code:

```
    if new_resource.peer_owner_id.nil?
      options[:peer_vpc_id] = new_resource.peer_vpc
      options = AWSResource.lookup_options(options, resource: new_resource)
    else
      options = AWSResource.lookup_options(options, resource: new_resource)
      options[:peer_vpc_id] = new_resource.peer_vpc
      options[:peer_owner_id] = new_resource.peer_owner_id
    end
```

So it will only lookup if you don't set the peer_owner_id (meaning you own the other VPC). Even if it lookup, you can set the specific ID. Its for that reason that lookup works, right? You can put the id or an object and it will identify what you want.

tyler-ball:
Hey y'all! I wrote a test to demonstrate this error, but cannot recreate it. Here is my test:

```
with_aws "with 2 VPCs not owned by Chef Provisioning" do
  vpc1 = nil
  vpc2 = nil
  before do
    converge do
      vpc1 = driver.ec2_resource.create_vpc(cidr_block: '10.0.0.0/24')
      vpc2 = driver.ec2_resource.create_vpc(cidr_block: '11.0.0.0/24')
      vpc1.wait_until {|v| v.state == 'available'}
      vpc2.wait_until {|v| v.state == 'available'}
    end
  end

  after do
    converge do
      vpc1.delete
      vpc2.delete
    end
  end

  it "aws_peering_connection 'test_peering_connection' with minimal parameters creates a active connection" do
    expect_recipe {
      aws_vpc_peering_connection 'test_peering_connection' do
        vpc vpc1.id
        peer_vpc vpc2.id
      end
    }.to create_an_aws_vpc_peering_connection('test_peering_connection',
      :'requester_vpc_info.vpc_id' => vpc1.id,
      :'accepter_vpc_info.vpc_id' => vpc2.id,
      :'status.code' => 'active'
    ).and be_idempotent
  end
end
```

What am I doing wrong that I cannot recreate this error?

cgunasree08:

PR #430, opened 2016-01-13
use ssh_gateway as defined in machine_options

kplimack:
this is my fix to (https://github.com/chef/chef-provisioning/issues/481), but there is probably a cleaner way to do this for someone who knows the guts and internals of provisioning....

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -1028,6 +1028,8 @@ EOD
       #Enable pty by default
       options[:ssh_pty_enable] = true
       options[:ssh_gateway] = machine_spec.reference['ssh_gateway'] if machine_spec.reference.has_key?('ssh_gateway')
+      options[:ssh_gateway] = machine_options[:machine_options][:ssh_gateway] if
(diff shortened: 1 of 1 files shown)


tyler-ball:
I'm not sure how `ssh_gateway` was not getting persisted into the machine_spec. It gets stored into the `machine_spec` when [creating the instance](https://github.com/chef/chef-provisioning-aws/blob/63757067a4128e9212ab9189078d5476df3bb6db/lib/chef/provisioning/aws_driver/driver.rb#L1480).

I [added a test](https://github.com/chef/chef-provisioning-aws/commit/3cc3c87892c490e2f63fc028159ba007de1d741d) configuring `ssh_username` and `ssh_options` with [how they are documented](https://github.com/chef/chef-provisioning/blob/d65512aea02d872ea155c7a0337959b4d0cdc84e/README.md#machine-options). The test does not include `ssh_gateway` because I do not know how to create one. But when I was testing it out locally `ssh_gateway` was being set on the `Net::SSH` object.

I'm going to close this for now because I don't know how you were getting `machine_options[:machine_options]` populated. Please re-open if you are still seeing this issue and we can troubleshoot more!

PR #434, opened 2016-01-15
Added aws_nat_gateway resource and provider:

joaogbcravo:
Added nat gateway provider that will enable the management by chef of http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-nat-gateway.html .
The creation and destroy waits for the the state 'available' and 'deleted' to make sure that it was really created/destroyed.

- Added a new route type to route table. Now routes can be pointed to nat_gateway_id.
- On VPC purge, it will delete nat_gateways in the process.

Note: Unfortunately, at the time, there isn't yet Aws::EC2::NatGateway class, so one was created to simulate is behauviour using ec2_client.


jgoulah:
👍    works for me (only tested create and idempotence)

tyler-ball:
Closing this in favor of #482

Issue #317, opened 2015-09-09
powershell user_data should support winrm over ssl by default


hh:
Currently we open 5986 (winrm ssl) but do not enable it.

``` ruby
    def user_data
      # TODO: Make this use HTTPS at some point.
      <<EOD
<powershell>
winrm quickconfig -q
winrm set winrm/config/winrs '@{MaxMemoryPerShellMB="300"}'
winrm set winrm/config '@{MaxTimeoutms="1800000"}'
winrm set winrm/config/service '@{AllowUnencrypted="true"}'
winrm set winrm/config/service/auth '@{Basic="true"}'

netsh advfirewall firewall add rule name="WinRM 5985" protocol=TCP dir=in localport=5985 action=allow
netsh advfirewall firewall add rule name="WinRM 5986" protocol=TCP dir=in localport=5986 action=allow

net stop winrm
sc config winrm start=auto
net start winrm
</powershell>
EOD
    end
```

We should support verifying the SSL Certificate before connecting.
The easiest way to do that would be to retrieve the RDP ssl certificate signature in via [GetConsoleOutput](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetConsoleOutput.html) as it's available in the console output on lines with RDPCERTIFICATE:

```
2015/09/09 13:54:11Z: EC2ConfigMonitorState: 0
2015/09/09 13:54:12Z: Windows sysprep configuration complete.
2015/09/09 13:54:14Z: AMI Origin Version: 2015.08.12
2015/09/09 13:54:14Z: AMI Origin Name: Windows_Server-2012-R2_RTM-English-64Bit-Base
2015/09/09 13:54:14Z: OS: Microsoft Windows NT 6.3.9600
2015/09/09 13:54:14Z: OsVersion: 6.3
2015/09/09 13:54:14Z: OsProductName: Windows Server 2012 R2 Standard
2015/09/09 13:54:14Z: OsBuildLabEx: 9600.17936.amd64fre.winblue_ltsb.150715-0840
2015/09/09 13:54:14Z: Language: en-US
2015/09/09 13:54:14Z: EC2 Agent: Ec2Config service v3.8.294
2015/09/09 13:54:16Z: Message: Waiting for meta-data accessibility...
2015/09/09 13:54:16Z: Message: Meta-data is now available.
2015/09/09 13:54:17Z: AMI-ID: ami-4dbcb67d
[...]

PR #437, opened 2016-01-28
Add support for ssl user_data provisioning fixing #317

hh:
When using winrm_transport ssl, reuse the RDP certificate for use with
winrm so we can capture the ssl fingerprint from the aws console.


hh:
FIxes #317

Issue #438, opened 2016-01-28
When instance.console_output is large, windows never becomes ready


hh:
Due to some type of bug in instance.console_output.output, if it's only called once, the AWS service or underlying gem may not have retrieved all the log yet, and it consisently
is unable to read the last lines which often include the latest RDP cert or the output saying windows is ready to use.

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #441, opened 2016-01-28
Winrm+RDP/SSL Fingerprint and Encrypted Password Caching

### LINKED ISSUE #348: support winrm over https transport

hh:
Fixes  #348, has been waiting on underlying changes to;
chef-provisioning and WinRM

#438 and probably a few others.

Thanks for all the help

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -987,10 +987,27 @@ EOD
 
     def create_winrm_transport(machine_spec, machine_options, instance)
       remote_host = determine_remote_host(machine_spec, instance)
+      username = machine_options[:winrm_username] || 'Administrator'
+      # default to http for now, should upgrade to https when knife support self-signed
+      transport_type = machine_options[:winrm_transport] || 'http'
+      type = case transport_type
+             when 'http'
+               :plaintext
+             when 'https'
+               :ssl
+             end
+      if machine_spec.reference[:winrm_port]
+        port = machine_spec.reference[:winrm_port]
+      else #default port
+        port = case transport_type
+               when 'http'
+                 '5985'
+               when 'https'
+                 '5986'
+               end
+      end
+      endpoint = "#{transport_type}://#{remote_host}:#{port}/wsman"
 
-      port = machine_spec.reference['winrm_port'] || 5985
-      endpoint = "http://#{remote_host}:#{port}/wsman"
-      type = :plaintext
       pem_bytes = get_private_key(instance.key_name)
 
       # TODO plaintext password = bad
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -648,11 +648,14 @@ EOD
       # sending out encrypted password, restarting instance, etc.
       if machine_spec.reference['is_windows']
         wait_until_machine(action_handler, machine_spec, "receive 'Windows is ready' message from the AWS console", instance) { |instance|
-          output = instance.console_output.output
-          if output.nil? || output.empty?
+          instance.console_output.output
+          # seems to be a bug as we need to run this twice
+          # to consistently ensure the output is fully pulled
+          encoded_output = instance.console_output.output
+          if encoded_output.nil? || encoded_output.empty?
             false
           else
-            output = Base64.decode64(output)
+            output = Base64.decode64(encoded_output)
             output =~ /Message: Windows is Ready to use/
           end
         }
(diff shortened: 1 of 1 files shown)


lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -984,31 +1033,115 @@ def default_ami_for_region(region, criteria = {})
...
-      type = :plaintext
       pem_bytes = get_private_key(instance.key_name)
 
-      # TODO plaintext password = bad
-      password = machine_spec.reference['winrm_password']
-      if password.nil? || password.empty?
-        encrypted_admin_password = instance.password_data.password_data
-        if encrypted_admin_password.nil? || encrypted_admin_password.empty?
-          raise "You did not specify winrm_password in the machine options and no encrytpted password could be fetched from the instance"
+      if machine_options[:winrm_password]
```
tyler-ball: Is `winrm_password` still being stored in the machine_spec?  Did you pull it out of there for security reasons?

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -984,31 +1033,115 @@ def default_ami_for_region(region, criteria = {})
...
+          decoded = Base64.decode64(encrypted_admin_password)
         end
-        decoded = Base64.decode64(encrypted_admin_password)
-        private_key = OpenSSL::PKey::RSA.new(pem_bytes)
+        # decrypt so we can utilize
+        private_key = OpenSSL::PKey::RSA.new(get_private_key(instance.key_name))
         password = private_key.private_decrypt decoded
       end
 
+      disable_sspi =  machine_options[:winrm_disable_sspi] || false # default to Negotiate
```
tyler-ball: Generally we want to store things like connection config into the `machine_spec`.  The reason is so that if you have code like

```ruby
machine 'windows_machine' do
  machine_options {...
    :winrm_no_ssl_peer_verification => true
  }
  action :ready
end
```

Then you later want call the same resource with `action :converge`, you don't need to re-specify the `winrm_no_ssl_peer_verification` config

tyler-ball: Or am I totally wrong and you get that from the stored `winrm_ssl_thumbprint` later?

tyler-ball:
Some questions about storing items in the `machine_spec` vs `machine_reference` but otherwise LGTM

lib/chef/provisioning/aws_driver/driver.rb
hh: Storing the clear text password in the node data.... it's ok to store it in the machine_options, but I wouldn't want that to persist within the machine_spec

lib/chef/provisioning/aws_driver/driver.rb
hh: We store the  winrm_ssl_thumbprint within the machine_reference the first time it's verified, and pull it from there afterwards.

hh: Since we want to be able to connect using machine_options :winrm_password and :winrm_disable_sspi etc so that we could use action :converge later

hh:
:+1:

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: I agree - we shouldn't store it in the machine_spec

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -984,31 +1033,115 @@ def default_ami_for_region(region, criteria = {})
...
+        basic_auth_only: basic_auth_only,
+        no_ssl_peer_verification: no_ssl_peer_verification,
       }
 
+      if no_ssl_peer_verification or type != :ssl
+        # =>  we won't verify certs at all
+        Chef::Log.warn "No SSL or no peer verification"
+      elsif machine_spec.reference['winrm_ssl_thumbprint']
+        # we have stored the cert
+        Chef::Log.warn "Using stored fingerprint"
```
tyler-ball: Should this really be a warning to the user?  Seems to me like it should be info

lib/chef/provisioning/aws_driver/driver.rb
tyler-ball: Currently, we don't overwrite the transport options once they have been set the first time.  For example you could create a `machine 'my_machine'` and set the `ssl_username` to 'foo' and that is the username we will use for all later occurances of `machine 'my_machine'`.

So for now, lets not try and change that.  If you initially have your machine_options setup to use SSL with thumbprint auth, then thats how it will stay.

tyler-ball:
One more question about a log level then I'll merge this.  Thanks @hh!

hh:
@tyler-ball looks like we should be good here now

tyler-ball:
Thanks for the patience and the contribution @hh!

Issue #452, opened 2016-02-04
When creating security-groups, the name should not be case-sensitive


dreamnite:
If I have a security group that was previously created with the name "Abc" and try:
```ruby
aws_security_group "ABC" do 
...
end
```
It will fail with an error "AWS::EC2::Errors::InvalidGroup::Duplicate". Since Amazon is treating the item as case-insensitive, the driver should as well.

tyler-ball:
Closed by https://github.com/chef/chef-provisioning-aws/issues/457

tyler-ball:
Actually, I'm reverting the fix I merged in #457. Aws has no case-insensitive way we can query security groups, so there is no way for us to tell from a resource named `"ABC"` whether there is a different case of that same security group that already exists. We could try every permutation on casing, but that is a bad idea.

Thats also the same error that you get if you try to create `"ABC"` then `"Abc"` through the AWS console. So I'm unfortunately going to close this as won't fix.

PR #457, opened 2016-02-10
make sure security group name is lowercase

scalp42:
A shot at https://github.com/chef/chef-provisioning-aws/issues/452


tyler-ball:
Chef Resources are case sensitive, so this seems a good enough solution to me.

For most AWS objects the name is just a tag, so I don't see this being an issue for other objects except ones that are also keyed off their name (instead of a randomly generated id).

Thanks @scalp42 !

tyler-ball:
Actually, reverting this PR - see https://github.com/chef/chef-provisioning-aws/issues/452#issuecomment-217021543

Thanks for the attempt though!

PR #483, opened 2016-05-11
fix how machine_options are specified

jgoulah:
attempt to fix how machine_options is specified in the README, yes the previous incarnation works, but fails after trying to add more machine_options which leads to confusing "bugs" like this (https://github.com/chef/chef-provisioning/issues/481) where the other options don't get picked up


nathenharvey:
👍

PR #504, opened 2016-09-30
Prioritize usage of key_data of user provisioning config over all private key settings

baurzhan-konurbayev:
Sometimes user might not have a private key of instance's key pair, but one does have SSH access and private key of root or any other legit user. So, in this case it would be handy to fill :key_data of :ssh_options iin user config and work this way. 

The problem arises when you don't have a key pair. If you try to converge you will get a message like:
> Server has key name '#{instance.key_name}', but the corresponding private key was not found locally

This pull request was inspired by chef-provisioning-fog, since it works exactly like in PR.

Thanks.

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -1210,6 +1210,31 @@ EOD
       end
     end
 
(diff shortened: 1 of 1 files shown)


tyler-ball:
Thanks for the PR! I added some test coverage and merged

Issue #306, opened 2015-08-31
machine_batch should batch requests


tyler-ball:
If a user has a recipe containing the following:

``` ruby
machine_batch "apache cluster" do
  (1..10).each do |i|
    machine "apache_#{i}" do
      machine_options #...
      run_list ["recipe[apache]"]
    end
  end
end
```

We should be only be making 1 request to AWS to create all 10 machines.  With the AWS SDK V2 we can set a [`min_count` and `max_count`](https://github.com/chef/chef-provisioning-aws/blob/64e23b6acae83a0417f06795b6c4b8812635e71a/lib/chef/provisioning/aws_driver/driver.rb#L678) parameter on the `create_instances` call.

charlesjohnson:
See also #17, #227

bpoweski:
This is useful for properly utilizing placement groups.

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

Issue #17, opened 2014-11-07
Support parallel creation with one API call


jkeiser:
allocate_machines, ready_machines!

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #510, opened 2016-11-05
Create multiple instances using one call when using machine_batch (#306)

bpoweski:
Prior to this change each EC2 instance was created by making a separate `#create_instances` call.  This approach is preferred when utilizing placement groups per Amazon's recommendations below:

> First, you create a placement group and then you launch multiple instances into the placement group. We recommend that you launch the number of instances that you need in the placement group in a single launch request and that you use the same instance type for all instances in the placement group. If you try to add more instances to the placement group later, or if you try to launch more than one instance type in the placement group, you increase your chances of getting an insufficient capacity error.
 
I attempted to run the integration tests with a live EC2 accounts before making these changes, unfortunately many of them failed sporadically due to what appears to race conditions around the duration of provisioning.

Related #17


bpoweski:
Any plans on merging?

cgunasree08:

PR #513, opened 2016-11-21
Workaround S3 endopoints in routes

adecarolis:
This patch prevents chef-provisioning from crashing when a route table includes a S3 endpoint.


tyler-ball:
I have no idea how to replicate this for testing setup, but it looks like a good catch to have. I am surprised that aws would put `nil` into the array they return, but ¯\_(ツ)_/¯

PR #515, opened 2016-12-02
Add support for RDS creation from DB snapshot

GolubevV:
Currently, in the chef-provision-aws there is no possibility to create AWS RDS instance, basing on existing DB snapshot - they new instances are always expected to be created from scratch.

This PR is targeted to provide support for creating RDS instances from the snapshot if it was specified as parameter for the resource. It also updates the minimal required version of AWS-SDK Ruby gem in Gem spec as default value (2.1.26) is no longer compatible with chef-provisioning-aws 1.11.0 and higher.

The code has been tested and is already used for provisioning our current instances of RDS since last 3-4 month.


tyler-ball:
This looks great @GolubevV ! I'm going to try and write some tests for this but I should be able to merge it

PR #519, opened 2017-01-04
Fix Classic use of SG names in rules

dafyddcrosby:
Using security group names lead to an issue which lead to an incorrect
intersection. This was not a problem with VPC security groups, but was
with Classic security groups, giving a 'duplicate security group' error.

Signed-off-by: David Crosby <dtcrsby@gmail.com>


tyler-ball:
Thanks @dafyddcrosby ! I'm not very familiar with the classic setup so I appreciate this PR. The test suite passed

PR #523, opened 2017-02-07
aws_rds_instance - update to support modify, and wait for create

keen99:
TODO:  
* merge master and update to support restore from snapshot functionality.
* update documentation
* include examples

Incomplete due to changes in master, but otherwise working for me in an older fork.   

Previously, rds was create-only, no modifications would apply.   this adds logic to handle update.   Also added options around waiting for update/create/delete to complete, and logic to wait for create.  

Note that updates (modify) are -always- run on every convergence, because we cannot sanely compare current to desired state.   In reality, this means that the master password is 'updated' every run - but no other state changes actually apply unless they need to.   

Users will definitely want to review their `additional_options( apply_immediately: ... )` state - this defaults to applying during the maint window.


cgunasree08:

PR #522, opened 2017-01-30
Fixed the credentials file environment variable name.

msonnabaum:
The AWS_CREDENTIAL_FILE environment variable isn't the right variable to load a custom path to a credentials file. The correct variable is AWS_SHARED_CREDENTIALS_FILE.

See the discussion here for more background: https://github.com/aws/aws-cli/issues/1037

I retained support for AWS_CREDENTIAL_FILE so that this doesn't introduce a backwards incompatible change.

PR #524, opened 2017-02-07
Fixes bug preventing disassociating eips.

msonnabaum:
Without this fix, the disassociating an eip like so:

```ruby
  aws_eip_address "some-eip" do
    machine false
  end
``` 

Results in:

```
    NameError
    ---------
    No resource, method, or local variable named `aws_object' for `Chef::Provider::AwsEipAddress ""'
```


tyler-ball:
Same comment from [here](https://github.com/chef/chef-provisioning-aws/pull/522#issuecomment-280060180)

msonnabaum:
Done

PR #527, opened 2017-03-06
update aws_cloudwatch_alarm to support float thresholds

keen99:
thresholds for cw alarms are often floats, not just integers.  the upstream interfaces support then, so just accept either float or int and pass them.


tyler-ball:
Thanks @keen99 !

PR #530, opened 2017-04-17
[Do not merge]Update chef-provisioning-aws to use AWS SDK v2

harikesh-kolekar:
Currently, the Chef Provisioning AWS gem depends on v1 of the AWS SDK gem, a gem which does not support JSON v2. Many of the ChefDK dependencies are now moving to JSON v2, which is going to cause an increase in the number of gem conflicts that will occur until a point where we can no longer build the ChefDK.


lib/chef/provider/aws_s3_bucket.rb
```diff
@@ -74,13 +72,27 @@ def destroy_aws_object(bucket)
 
   private
 
+  def website_exist?(new_resource,bucket)
+    return true if new_resource.driver.s3.get_bucket_website(bucket: bucket.name)
+  rescue
```
btm: in principle, blind rescues should be avoided, especially without any logging. Is there a specific error we can rescue here?

Also, we're only returning false if we rescue and returning nil if an exception isn't raised. Are we sure that's never going to happen?

lib/chef/provider/aws_s3_bucket.rb
```diff
@@ -74,13 +72,27 @@ def destroy_aws_object(bucket)
...
+    return false
+  end
+
+  def create_website(bucket,new_resource )
+    website_configuration = Aws::S3::Types::WebsiteConfiguration.new(
+            new_resource.website_options)
+    s3 = new_resource.driver.s3
+    s3.put_bucket_website( bucket: bucket.name,  website_configuration:website_configuration)
+
```
btm: there's a blank line here and the whitespace in the line above this is inconsistent (remove the space between `(` and `bucket:`, I think you want to have a single space after the colon between keywords, i.e. `bucket: bucket.name` and `website_configuration: website_configuration`.

lib/chef/provider/aws_dhcp_options.rb
```diff
@@ -15,7 +15,7 @@ def create_aws_object
 
     converge_by "create DHCP options #{new_resource.name} in #{region}" do
       dhcp_options = new_resource.driver.ec2.dhcp_options.create(options)
-      retry_with_backoff(AWS::EC2::Errors::InvalidDhcpOptionsID::NotFound) do
+      retry_with_backoff(Aws::EC2::Errors::InvalidDhcpOptionsID::NotFound) do
```
tyler-ball: A recommendation: Whenever switching from `AWS` to `Aws` we should always prepend it with `::` - so this would be `::Aws::EC2::Errors::InvalidDhcpOptionsID::NotFound`.

AWS generates their classes on the fly (last I checked) so there is no file we can require. Without requiring the file that exposes these classes, we need to make sure Ruby knows we are referencing the base `Aws` module and not a sub-module of the current location.

lib/chef/provider/aws_s3_bucket.rb
```diff
@@ -26,21 +25,19 @@ def action_create
...
-            new_resource.website_options)
+          create_website(bucket,new_resource )
         end
       end
     else
-      if bucket.website?
+      if website_exist?(new_resource,bucket)
         converge_by "disable website configuration for bucket #{new_resource.name}" do
-          bucket.website_configuration = nil
+          new_resource.driver.s3.delete_bucket_website(bucket: bucket.name)
```
tyler-ball: To use the new V2 library this should be `s3_client` instead of `s3`

lib/chef/provisioning/aws_driver/driver.rb
```diff
@@ -764,12 +763,12 @@ def destroy_machine(action_handler, machine_spec, machine_options)
...
     def cloudsearch(api_version="20130101")
       @cloudsearch ||= {}
-      @cloudsearch[api_version] ||= AWS::CloudSearch::Client.const_get("V#{api_version}").new
+      @cloudsearch[api_version] ||= Aws::CloudSearch::Client.const_get("V#{api_version}").new
       @cloudsearch[api_version]
     end
 
     def ec2
-      @ec2 ||= AWS::EC2.new(config: aws_config)
+      @ec2 ||= Aws::EC2.new(config: aws_config)
```
tyler-ball: These methods (`ec2`, elb`, `elasticache`, etc.) which previously defined the clients have been replaced by the dynamically defined methods on [line 744](https://github.com/MsysTechnologiesllc/chef-provisioning-aws/blob/c5d17253559d293b0087bb49075636c5d0c97a63/lib/chef/provisioning/aws_driver/driver.rb#L774). These old methods can be removed and instead the methods like `ec2_client`, `ec2_resource`, `s3_client`, `s3_resource`, etc. should be used

harikesh-kolekar:
*Do not merge with master*
The code migration from AWS SDK V1 to V2  is done for the S3 resource, the other resources are not migrated and tested.

jjlimepoint:
I added a PR against the source repo for this, to make EC2 instances work as well - if I should PR it against this instead, I can do that

PR #532, opened 2017-04-28
correctly symbolize bootstrap_options, making chef13 go again

jjlimepoint:
correctly symbolize bootstrap_options, making chef13 go again - this one was missed in the symbolizing!

This set of changes allows current chef13/cheffish13 to correctly provision AWS machines

### DIFF
--- a/lib/chef/provisioning/aws_driver/aws_provider.rb
+++ b/lib/chef/provisioning/aws_driver/aws_provider.rb
@@ -29,7 +29,7 @@ class AWSProvider < Chef::Provider::LWRPBase
   end
 
   def region
-    new_resource.driver.aws_config.region
+    new_resource.driver.aws_config[:region]
   end
 
   #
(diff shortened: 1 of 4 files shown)


jjlimepoint:
Supposedly this specifically affects using machine_batch - it works without this work machine on its own, but not machine_batch

dreamnite:
This works for me with one small issue: the deprecated ith the deprecated `use_private_ip_for_ssh: true`. When it converts the value into `transport_address_location: :private_ip` it gives an error:

 `  Unexpected method []= for MergedConfig with arguments [:transport_address_location, :private_ip]`

When left off, or set with transport_address_location it works fine.

BenLiyanage:
This seems to work similarly to #535, but with more code.  Is the other way better for dealing with this issue?

jjlimepoint:
The reason for more code is that i needed to update deep_symbolize to work with lists, or else things like block_device_mapping don't actually work right, since they are lists that contain hashes, which cheffish also messes with.

tyler-ball:
Question: what issues did you encounter with `aws_config.region`? With your code I get the error `undefined method `[]' for <AWS::Core::Configuration>:AWS::Core::Configuration` trying to run the aws_auto_scaling_group_spec

jjlimepoint:
I was getting nil returns with :region, but correct returns with .region - i assume the object changed to an actual object, at some point....

Issue #533, opened 2017-06-03
machine_batch fails while machine alone resouce works fine


BGmot:
I think I am missing something very basic here and any help would be appreciated.
The code below woks without any problem:
```
require 'chef/provisioning/aws_driver'
with_driver 'aws'
machine 'ey-test' do
  machine_options( {
    bootstrap_options: {
      associate_public_ip_address: false
    }
  } )
end
```

Following code fails (obviously I'd like to spin up more machines but even this simplistic test fails)
```
require 'chef/provisioning/aws_driver'
with_driver 'aws'
machine_batch do
  machine 'ey-test' do
    machine_options(
      { bootstrap_options: {
          associate_public_ip_address: false
        }
      }
    )
  end
end
```
chef-client's log:
```
  * machine_batch[default] action converge[2017-06-03T17:03:33-04:00] INFO: Processing machine_batch[default] action converge (kafka-cluster::test-aws line 3)
/opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/cheffish-5.0.1/lib/cheffish/merged_config.rb:6: warning: toplevel constant Mash referenced by Chef::Mash
/opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/cheffish-5.0.1/lib/cheffish/merged_config.rb:7: warning: toplevel constant Mash referenced by Chef::Mash
[2017-06-03T17:03:46-04:00] INFO: [Aws::EC2::Client 200 0.650918 0 retries] describe_instances(instance_ids:["i-06e40b1221bee104c"])

[2017-06-03T17:03:46-04:00] INFO: [Aws::EC2::Client 200 0.155659 0 retries] describe_instances(instance_ids:["i-06e40b1221bee104c"])

[2017-06-03T17:03:46-04:00] WARN: Machine ey-test (i-06e40b1221bee104c) is terminated.  Recreating ...
[2017-06-03T17:03:46-04:00] INFO: Processing aws_key_pair[chef_default] action create (basic_chef_client::block line 1385)
[2017-06-03T17:03:47-04:00] INFO: [AWS EC2 200 0.541622 0 retries] describe_key_pairs(:filters=>[{:name=>"key-name",:values=>["chef_default"]}])

[2017-06-03T17:03:47-04:00] INFO: [AWS EC2 200 0.252683 0 retries] describe_key_pairs(:key_names=>["chef_default"])

[...]

PR #535, opened 2017-06-08
Fix machine_batch

BGmot:
Fixes issue https://github.com/chef/chef-provisioning-aws/issues/533


jjlimepoint:
looks like basically the same fix as the one in PR #532 , however you probably also want something like  https://github.com/chef/chef-provisioning-aws/pull/532/commits/3bdca320298a77862601eb044af5a4584fb15fa9 or else lists won't be correctly symbolized, causing weird failures in disk/network attachments

tyler-ball:
Closing in favor of #532

PR #542, opened 2017-10-23
MSYS-669: Migrate aws_rds_parameter_group  from AWS SDK V1 to V2

piyushawasthi:
Signed-off-by: piyushawasthi <piyush.awasthi@msystechnologies.com>

### DIFF
--- a/lib/chef/resource/aws_rds_parameter_group.rb
+++ b/lib/chef/resource/aws_rds_parameter_group.rb
@@ -24,7 +24,7 @@ class Chef::Resource::AwsRdsParameterGroup < Chef::Provisioning::AWSDriver::AWSR
       parameters += more_results[:parameters]
       marker = more_results[:marker]
     end
-    # object[:parameters] = parameters
+    driver.rds.reset_db_parameter_group(db_parameter_group_name: name, parameters: parameters)
 
     object
   rescue ::Aws::RDS::Errors::DBParameterGroupNotFound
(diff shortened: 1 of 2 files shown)


spec/integration/aws_rds_parameter_group_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'pry'
```
btm: let's not leave pry here please

lib/chef/resource/aws_rds_parameter_group.rb
```diff
@@ -24,7 +24,7 @@ def aws_object
       parameters += more_results[:parameters]
       marker = more_results[:marker]
     end
-    # object[:parameters] = parameters
+    driver.rds.reset_db_parameter_group(db_parameter_group_name: name, parameters: parameters)
```
tyler-ball: What is the reason for doing this in the `aws_object` method? I would think something like this would occur in the provider's update logic.

spec/integration/aws_rds_parameter_group_spec.rb
piyushawasthi: Sorry for that, It'll never happen again.

lib/chef/resource/aws_rds_parameter_group.rb
piyushawasthi: @tyler-ball update limit is 20 that's why I am using `reset_db_parameter_group` here this is pretty faster to reset all parameters.

PR #546, opened 2017-10-26
Fix for failing specs

Vasu1105:
This PR contains fix for failing specs for load balancer and route table. Also fixed regular expression which was used to match instance id.

Signed-off-by: Vasu1105 <vasundhara.jagdale@msystechnologies.com>


Vasu1105:
@tyler-ball #545 fix

PR #548, opened 2017-11-02
[MSYS-698] Empty tags are not working

harikesh-kolekar:
Signed-off-by: Harikesh Kolekar <harikesh.kolekar@msystechnologies.com>


btm:
@harikesh-kolekar is there a test we can add here to make sure that empty tags works?

harikesh-kolekar:
@btm https://github.com/chef/chef-provisioning-aws/blob/master/spec/integration/aws_dhcp_options_spec.rb#L58 **and**
https://github.com/chef/chef-provisioning-aws/blob/master/spec/integration/aws_auto_scaling_group_spec.rb#L105
The above integration specs will check `removes all tags` test case.

btm:
@harikesh-kolekar okay. And the integration tests don't run in Travis? We have to run them manually?

PR #549, opened 2017-11-06
[MSYS-700] Migrate aws_cloudsearch_domain from v1 to v2

Vasu1105:
Note: @tyler-ball  @btm Added deperication warning for  cloudsearch_api_version attribute since there is no support for this in version 2 Ref http://docs.aws.amazon.com/sdkforruby/api/Aws/CloudSearch/Client.html

Signed-off-by: Vasu1105 <vasundhara.jagdale@msystechnologies.com>

### DIFF
--- a/lib/chef/provider/aws_cloudsearch_domain.rb
+++ b/lib/chef/provider/aws_cloudsearch_domain.rb
@@ -177,6 +177,6 @@ class Chef::Provider::AwsCloudsearchDomain < Chef::Provisioning::AWSDriver::AWSP
   end
 
   def cs_client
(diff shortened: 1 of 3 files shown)


NimishaS:
Please note that this can be a breaking change since we have to remove `cloudsearch_api_version` attribute which was previously supported by V1 API, but isn't supported by V2 API.

thommay:
Might be better to keep the attribute and just emit a deprecation warning if it's used (and throw away anything passed to it).

Vasu1105:
@NimishaS @thommay Done necessary changes for deprecation warning for cloudsearch_api_version attribute.

PR #552, opened 2017-11-08
Fix load balancer test failures

tyler-ball:
@Vasu1105 I think this is ready for merge!

I reverted https://github.com/chef/chef-provisioning-aws/pull/552/commits/a464e484611228789c41b46bef441fe4c2760945 because that change effectively makes the `. create_an_aws_load_balancer` compare the exact same two values.

The first string passed to `create_an_aws_load_balancer` is used to lookup an aws_object based on the [aws_object method](https://github.com/chef/chef-provisioning-aws/blob/f11a11d6f7762e60a653b6794309fe89384f13bc/lib/chef/resource/aws_load_balancer.rb#L18). With the change you had, it was then comparing that object to another object looked up from AWS in the exact same way. So the expected and actual values were the same looked up object from AWS.

I fixed some of the comparison operator's failure to deal with Structs and I think this is ready for review now!

### DIFF
--- a/lib/chef/provisioning/aws_driver/driver.rb
+++ b/lib/chef/provisioning/aws_driver/driver.rb
@@ -616,7 +616,8 @@ module AWSDriver
 
         if instance_ids_to_remove.size > 0
           perform_action.call("  remove instances #{instance_ids_to_remove}") do
-            actual_elb.instances.remove(instance_ids_to_remove)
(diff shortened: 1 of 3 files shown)


btm:
@NimishaS @Vasu1105 I think we want/need to get this merged before final release. Can you take a look?

Vasu1105:
LGTM @btm

NimishaS:
LGTM

PR #553, opened 2017-11-09
[MSYS-705] Adding integration testcases for aws_cache_cluster resource

NimishaS:
`destroy` method is called immediately after the integration test finishes for any resource. `aws_cache_cluster` resource takes around 10 minutes to become available and it can't be deleted till it's not in available state.
So this PR is adding `wait_for_cache_cluster_state` method for `aws_cache_cluster` resource, so that the integration testcase can run successfully.

Please note that this adds a delay of 10 minutes in the integration testcases and also while running any recipe of ` aws_cache_cluster` resource.

### DIFF
--- a/lib/chef/provider/aws_cache_cluster.rb
+++ b/lib/chef/provider/aws_cache_cluster.rb
@@ -1,14 +1,22 @@
 require 'chef/provisioning/aws_driver/aws_provider'
+require 'retryable'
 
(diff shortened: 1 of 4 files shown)


NimishaS:
@btm , @thommay , please comment if these changes are desirable. We had to add a delay of 10 minutes in `aws_cache_cluster` resource to make the integration testcase work.

thommay:
so if we need to do this we should disable these integration tests by default and hide them under a `slow` tag.

PR #554, opened 2017-11-16
Fix for bootstrap_options included in knife.rb

ndobson:
This bug was introduced in v2.2.0 here:
https://github.com/chef/chef-provisioning-aws/compare/v2.1.0...v2.2.0#diff-0410aad4b9403c10f07fbeb410833ef5L852


ndobson:
@btm

Issue #559, opened 2018-01-18
 machine/machine_image resources no longer wait for transport


bnutt:
chef-provisioning-aws: 3.0.0
chef-client: 13.6.4

After [change](https://github.com/chef/chef-provisioning-aws/commit/7c005f978e5b5f5df72ee9e8eebf20c9d84aa1cf#diff-0410aad4b9403c10f07fbeb410833ef5L1460), when a machine or machine_image calls wait_for_transport, it immediately skips  action_handler.report_progress "waiting for #{machine_spec.name} (#{instance.id} on #{driver_url}) to be connectable (transport up and running) ...". 

This makes spinning up a machine or machine_image unreliable because it no longer waits for SSH to be available and results in unavailable: network connection failed or broke: #<Errno::ECONNREFUSED: Connection refused - connect(2) for HOST> when attempting to transfer files to the newly provisioned machine.
 
This should be reverted to use transport.available? or there should be a fix for how to lookup the state of the instance so it properly checks if it is running. To workaround this as a compatibility patch, updating the file on the provisioner node to use transport.available? has worked to guarantee the the instance is warm and connectable.

PR #562, opened 2018-02-12
Transport available

jjlimepoint:
Since we were affected by https://github.com/chef/chef-provisioning-aws/issues/559  too (worked around with retries before, like a terrible human ;)), here's a patch that brings back actually waiting for SSH properly


tyler-ball:
Pulled into https://github.com/chef/chef-provisioning-aws/pull/564 - Thanks @jjlimepoint !

PR #575, opened 2018-09-07
replace driver with self.driver in get_driver_and_id

jjlimepoint:
this makes it work with latest chef.  to be honest, i am unsure how this ever worked - it should not have.  it relied on weird ruby scoping that does not work anymore


jjlimepoint:
Tested this on chef13 now, and it works there too.

